PR #71, opened 2013-08-26
fix DER encoding of BIT STRING values

siebertm:
Quote (http://luca.ntop.org/Teaching/Appunti/asn1.html):

>  In a primitive encoding, the first contents octet gives the number of bits by which the length of the bit string is less than the next multiple of eight (this is called the "number of unused bits")

This was not followed, leaving wrongly encoded DER (which caused me some headache to debug  till I found this :-) ) 

This PR fixes this (BTW tests are failing for current master?)

### DIFF
--- a/js/asn1.js
+++ b/js/asn1.js
@@ -428,6 +428,12 @@ asn1.toDer = function(obj) {
         value.putInt16(obj.value.charCodeAt(i));
       }
     }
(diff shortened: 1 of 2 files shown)


JrSchild:
Hi @siebertm,
I know this issue is fairly outdated. Has the functionality mentioned in this PR been implemented?

PR #131, opened 2014-06-18
script + support to generate npm module for browserify

mvhenten:
Hi, here's what I've got so far. The ./import script will re-write your code quite a bit, but that wasn't me ( it's esprima! ) I started out with the intend of having at least an "easy" path with keeping up to date with the project, but it got out of hand.


dlongley:
#456 made forge compatible with browserify.

PR #205, opened 2014-12-10
update README.md: fixing incompatibility with Java

ttruderung:
Hi,

The example of RSA-OAEP encryption/decryption with SHA-256 and MGF1-SHA1 is not compatible with Java's RSA/ECB/OAEPWithSHA-256AndMGF1Padding (although this is claimed in the commend), at least for the BC provider (I have not checked it with any other providers). 

When changed to MGF1-SHA256, it becomes compatible with RSA/ECB/OAEPWithSHA-256AndMGF1Padding (with BC), as claimed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -850,21 +850,23 @@ var decrypted = privateKey.decrypt(encrypted, 'RSA-OAEP', {
   md: forge.md.sha256.create()
 });
 
-// encrypt data with a public key using RSAES-OAEP/SHA-256/MGF1-SHA-1
-// compatible with Java's RSA/ECB/OAEPWithSHA-256AndMGF1Padding
+// encrypt data with a public key using RSAES-OAEP/SHA-256/MGF1-SHA-256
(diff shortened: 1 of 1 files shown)


dlongley:
I believe that there's an incompatibility between the BC provider and SunJCE in this case. The README should actually be updated to show how to use both and indicate which example to use with which provider. The original example works with SunJCE (at least I believe it was SunJCE that was used originally), and I assume your example works with BC.

dlongley:
http://bouncy-castle.1462172.n4.nabble.com/Interoperability-issue-with-SunJCE-OAEP-td4656157.html

zarfide:
I happened to stumble over this thread while looking at JCE/BC interop issues -- they are compatible but JCE requires some extra parameters -- see the updated BC thread.  Just FYI.

davidlehn:
Sorry, this got auto-closed due to branch rename.  See https://github.com/digitalbazaar/forge/pull/936.

PR #214, opened 2015-01-16
base64.decode can return an array that's too big

linuxwolf:
When using forge.util.base64.decode (without providing a pre-allocated buffer), the returned Uint8Array can be larger than the actual encoded data.


dlongley:
Thanks!

Issue #219, opened 2015-02-25
Can't get bytes during aes update


nekolab:
Hello
I'm porting an app which using openssl to chrome app. 
Now I decide to use forge as AES encryptor instead of Openssl.

In this app, it sending data using a loop like:

```
cipher = new aes(iv)
while read():
    cipher.update(readContent)
    send(cipher.getBytes())
```

But I found in forge, I can just get encrypted data after calling finish().
The question is after calling finish(), I can't update cipher anymore.
Any ideas can let me get bytes without calling finish or restart cipher after finish?

dlongley:
The cipher API in forge exposes a buffer called `output` that stores encrypted data as it is generated. You can read from this buffer as you call `cipher.update` and before you call `cipher.finish` without issue. However, not every call to `cipher.update` will generate output (eg: you pass it too few bytes to output the next cipher block), so if that's a problem for your `send` API, then just check the size of the output buffer before trying to grab bytes from it. When you're done reading and encrypting the content, call `cipher.finish` and then send over the last block in `cipher.output`. See below:

```js
// start the cipher
var cipher = forge.cipher.createCipher('AES-CBC', key);
cipher.start({iv: iv});

while(someCondition) {
  cipher.update(forge.util.createBuffer(someText, 'utf8'));
  if(cipher.output.length() > 0) {
    send(cipher.output.getBytes());
  }
}

// finish encrypting and send final data
cipher.finish();
if(cipher.output.length() > 0) {
  send(cipher.output.getBytes());
}
```

Also, note that forge 0.6.x returns a binary-encoded string when you call `.getBytes()` on a buffer. If your `send` API needs the data in a Uint8Array instead, do this:

```js
send(s2a(cipher.output.getBytes()));

function s2a(str) {
  var view = new Uint8Array(str.length);
  for(var i = 0, j = str.length; i < j; i++) {
    view[i] = str.charCodeAt(i);
  }
  return view;
}
```

meowlab:
Thanks for your great help and notice!
Seems it's not a good way to test a cipher by feeding only two bytes to it :)

meowlab:
@dlongley Sorry to mention you again...
I just deep into the protocol and found something I didn't express clearly.

The protocol like a tcp pipe, it receive a message from one side, encrypt it, when another side receive this message, it will decrypt it.

[...]

PR #224, opened 2015-03-04
Byte stream mode implementation for CFB, OFB and CTR

nekolab:
Please see #219 for more information.


meowlab:
0 0...Emm...
Sorry for my refactor reference noise....
I didn't realized that before...So sorry...

PR #226, opened 2015-03-09
fix AES decipher iv transformation from byte array to byte buffer

fujifish:
There's a bug in transformIV when the iv is given as a byte array. The resulting byte buffer is always empty. This fix transforms the iv byte array into the byte buffer correctly.

This fix is critical for production use! Please do your best to merge as soon as possible.


dlongley:
Thanks @fujifish! Released 0.6.21 with this fix.

PR #227, opened 2015-03-09
Specify module format as amd for jspm

frankier:
See https://github.com/jspm/registry/wiki/Configuring-Packages-for-jspm#format


dlongley:
Thanks @frankier!

PR #232, opened 2015-04-01
Check for window.location.search

duereg:
Very small tweak... check for window.location.search in case it doesn't exist. 

Coming across this case when faking browser availability.


dlongley:
Thanks @duereg!

PR #249, opened 2015-05-03
Generate public PEM from private key

mattcollier:
This makes it easier to share a public PEM derived from the private key.

Test is included.

Users can now store only the private key easily derive the public PEM on an as needed basis.


OR13:
This PR is ancient and has not comments. recommend closing.

PR #267, opened 2015-06-15
Fix certificates added to PKCS#7 SignedData

roysjosh:


dlongley:
Landed as 2964bd7240b6ad9cb32eb97941e0317d28a0ef37.

PR #268, opened 2015-06-15
Match OpenSSL output for enveloped data

roysjosh:
Do not include IMPLICIT ASN.1 elements in the output

### DIFF
--- a/lib/pkcs7.js
+++ b/lib/pkcs7.js
@@ -1114,10 +1114,8 @@ function _encryptedContentToAsn1(ec) {
         ec.parameter.getBytes())
     ]),
     // [0] EncryptedContent
-    asn1.create(asn1.Class.CONTEXT_SPECIFIC, 0, true, [
-      asn1.create(asn1.Class.UNIVERSAL, asn1.Type.OCTETSTRING, false,
-        ec.content.getBytes())
-    ])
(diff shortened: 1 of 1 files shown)


roysjosh:
Ping.  Do you need any more info?

davidlehn:
Sorry for the delay!  I'm guessing no one has taken the time to figure out if this change is ok and won't break other code.  Is there some spec text related to this?  Are there tests that could be written?

roysjosh:
https://tools.ietf.org/html/rfc2315#section-10.1

ASN.1 IMPLICIT tags don't include the tag of the child element.  See http://luca.ntop.org/Teaching/Appunti/asn1.html or google asn.1 implicit tagged object.

This is another good example:
https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-X.690-200811-S!!PDF-E&type=items

Also compare to OpenSSL's smime output.

Tests could probably be written.  I can add a few if that would help.  There's a comment in lib/pkcs7asn1.js about the output differing and accepting both types so I didn't mess with the capturing bits at all.  I think you should continue to be lenient on parsing but switch to the more correct output since there does seem to be some interpretation differences.

roysjosh:
I should mention that the motivation for this change is that the JSS library used by Red Hat's Dogtag certificate server only accepts the encoding lacking the implicit element's child tag.

PR #271, opened 2015-06-24
Optional verification of certificate validity date in certificate chain

robinrutten:
pki.verifyCertificateChain method now has optional parameter to skip validity period check. Furthermore, added verifyOptions to tls connection so we can skip this verification step when we setup a tls connection.

### DIFF
--- a/js/tls.js
+++ b/js/tls.js
@@ -3545,7 +3545,7 @@ tls.verifyCertificateChain = function(c, chain) {
         }
 
         return ret;
-      });
+      }, c.verifyOptions);
   } catch(ex) {
     // build tls error if not already customized
     var err = ex;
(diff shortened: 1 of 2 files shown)


angelos:
Any news about this request? Code looks fine to me, and it's a good addition for special use cases in TLS certificate verification.

dlongley:
@robinrutten and @angelos,

I'll look into this soon. Before accepting I want to see if there's an easier way to allow people to specify custom behavior like this without having to (eventually) create a laundry list of options that must be supported in the core TLS module.

angelos:
In the vein of "scratch your own itch," this is what works for us :)

However, I can see that it's not the most future proof API change. Two suggestions,
- rely on some list of mandatory validations, and use that as part of the API (I'm not sure whether that exists), or
- make errors more complete: in stead of throwing an error at the first sight of problems, continue all parts of the validation and create a more complete report in the error. This allows the user to pick and choose which problems he wants to ignore. This does, however, mean totally overhauling the validation code.

dlongley:
Replaced by #386.

PR #283, opened 2015-08-05
Optional parameter minSupportedVersion for tls.createConnection()

rachidbm:
Added optional paramter minSupportedVersion to tls.createConnection(). This parameter specifies which TLS protocol should be supported at minimum by the other side. 
We use this to explicitly disable TLS v1.0.

### DIFF
--- a/js/tls.js
+++ b/js/tls.js
@@ -1101,8 +1101,7 @@ tls.handleClientHello = function(c, record, length) {
       }
     }
     if(version === null) {
-      // TODO RBM: remove log
-      console.log('ERROR: Could not find TLS version compatible with ClientHello!');
+      throw new Error('Could not find TLS version compatible with version in Hello message.');
     }
     c.version = {major: version.major, minor: version.minor};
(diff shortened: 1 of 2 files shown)


dlongley:
Forge 0.7.x (which is presently being developed on this branch that needs renaming: https://github.com/digitalbazaar/forge/tree/buffer-based-api) already has a feature to be explicit about the version you want to use when connecting. It was first introduced here: e12da765b6a4e1f21caefc58c4c860c9a48b7778, be6a024b26b3a62f7f0d69d577b1e1b8eb6896e2 but may have a few other changes later down the line.

To get support for this in 0.6.x, a backport that cherry picks what's needed should be performed. If any additional checks are necessary, we can layer those on top.

dlongley:
There may be some differences in the semantics of 0.7.x's `options.version` and what's provided here, so a closer look may be required to either reconcile them or ensure all common use cases around selecting a version are supported.

dlongley:
Replaced by #387.

PR #289, opened 2015-08-19
CMS/PKCS#7 support for RSAES-OAEP

stesie:
This pull request is based on my previous pull request #288.

Goal of this change is to be able to create Enveloped-Data messages in CMS syntax (RFC 5652) using the RSA-OAEP encryption scheme. After all CMS is a successor of PKCS#7 v1.5 which probably justifies for a seperate module in forge. However I'm currently not willing to implement whole CMS stuff but just the Enveloped-Data thing like PKCS#7 with RSAES-OAEP ... which actually looks like PKCS#7, even has the version number untouched, solely a different RSAES.

... hence I decided *not* to create a new CMS module which effectively would be either a copy of PKCS#7 module (or some common subset) ... but instead augment the PKCS#7 module. This is latter module now allows to pass `algorithm` and `schemeOptions` options to its `addRecipient` method and handles them by considering changes from RFC 5652 + handling reverse direction.

Besides I changed the public API of MGF1 module so it has two new public attributes:
* `algorithm` set to "mgf1", so a MGF object tells its type (like the MD objects do)
* `md` allowing to access the MD injected with the `create` call

... both are needed so the PKCS#7 module can access the info that it has to encode into the CMS structure.

### DIFF
--- a/js/mgf1.js
+++ b/js/mgf1.js
@@ -23,6 +23,9 @@ var mgf1 = forge.mgf.mgf1 = forge.mgf1 = forge.mgf1 || {};
  */
 mgf1.create = function(md) {
   var mgf = {
+    algorithm: 'mgf1',
+    md: md,
+
     /**
      * Generate mask of specified length.
      *
--- a/js/pkcs1.js
+++ b/js/pkcs1.js
@@ -63,8 +63,7 @@ var pkcs1 = forge.pkcs1 = forge.pkcs1 || {};
  *          label an optional label to use.
  *          seed the seed to use.
  *          md the message digest object to use, undefined for SHA-1.
- *          mgf1 optional mgf1 parameters:
- *            md the message digest object to use for MGF1.
+ *          mgf the mask generation function to use, undefined for MDF1(md)
  *
  * @return the encoded message bytes.
  */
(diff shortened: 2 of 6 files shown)


dlongley:
Thanks, stesie, this is great.

It will take some time for me to review this and the MGF1 API changes -- I want to ensure they work with the direction [0.7.x](https://github.com/digitalbazaar/forge/tree/0.7.x) is headed.

wottpal:
hey @digitalbazaar, are there any chances on getting this one merged at some point? :) We are relying on this specific pr as a dependency in our `package.json` for a while now and this doesn't feel like a great solution.

Thank you so much!
Dennis

dlongley:
@wottpal, This PR needs to be rebased because it has conflicts. @stesie, are you able to rebase it? If not, @wottpal, could you submit a new PR that has the changes you want to see? Either way, only once the conflicts are resolved, it can be reviewed and potentially pulled in. Thanks!

stesie:
I'm myself no longer using this to be honest, and I don't have much time spare right now.  Hence I won't be able to do it soonish.

So if anyone else wants to step up, feel free

PR #370, opened 2016-03-16
Adds the ability to create a PBKDF2 hash with SHA-512 digest

stwiname:


dlongley:
@stwiname,

The SHA-512 module adds itself to `forge.md`, so that way you can leave it out of your project if you don't want to use it (see: https://github.com/digitalbazaar/forge/blob/2115c6d6a1accbaba30336d2fa92d9a9f50bbb6e/js/sha512.js#L17-L22). So I don't think we need that first change.

However, the additional tests are always welcome.

stwiname:
Is there something i need to do to initialize `sha-512` then? 

If I remove the changes to `md.js` then the added tests fail.

davidlehn:
sha512 support isn't added by default to the md module so that test will have to load it.  Look at the code at the bottom of the file and update it to load sha512.

dlongley:
Addressed by #456.

PR #388, opened 2016-04-25
Distinguished name length in certificate request

RoomUnitBDRThermea:
Made sure the pull request is now based on the latest 0.6.x release of Forge.

According to the TLS specification the handshake message certificate_request, sent as part of the server_hello messages, should include the length of each individual distinguished name before the distinguished name itself.

### DIFF
--- a/js/tls.js
+++ b/js/tls.js
@@ -3197,13 +3197,14 @@ tls.createCertificateRequest = function(c) {
   // common RSA certificate type
   certTypes.putByte(0x01);
 
-  // TODO: verify that this data format is correct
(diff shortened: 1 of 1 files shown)


davidlehn:
That was odd.  Random failure for iojs.  Rerun of tests and it worked.  I "fixed" that issue by bumping travis test node 4.x version to 4.4 and dropping iojs since it's no longer developed.

RoomUnitBDRThermea:
Well even in the hard science of software engineering, random stuff like this tends to happen sometimes :).
Thanks for the quick reply and for fixing it!

PR #389, opened 2016-04-25
Invalid empty tls messages

RoomUnitBDRThermea:
TLS messages of type handshake, alert or change_cipher_spec are not allowed to be empty.

### DIFF
--- a/js/tls.js
+++ b/js/tls.js
@@ -3377,8 +3377,8 @@ tls.queue = function(c, record) {
     return;
   }
 
-  if (record.fragment.length() == 0) {
-    if (record.type === tls.ContentType.handshake ||
+  if(record.fragment.length() === 0) {
+    if(record.type === tls.ContentType.handshake ||
       record.type === tls.ContentType.alert ||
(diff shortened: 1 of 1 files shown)


js/tls.js
```diff
@@ -3377,6 +3377,15 @@ tls.queue = function(c, record) {
     return;
   }
 
+  if (record.fragment.length() == 0) {
```
dlongley: Minor style nit, in this lib we don't put a space between "if" and parenthesis: `if (` => `if(`. We also always compare using triple equals.

js/tls.js
```diff
@@ -3377,6 +3377,15 @@ tls.queue = function(c, record) {
...
+    if (record.type === tls.ContentType.handshake ||
```
dlongley: Same minor style nit here.

dlongley:
@RoomUnitBDRThermea,

Could you fix the minor style nits and squash? Thanks!

dlongley:
Might need a rebase on master as well.

PR #403, opened 2016-06-03
UTF8 decode pki.RDN attributes

pk13nv:
When creating PKCS7 sign with certificate having DN with non-ascii characters, pkcs7 structure contains double encoded DN atributes (double UTF8 encoding)

### DIFF
--- a/js/x509.js
+++ b/js/x509.js
@@ -507,6 +507,9 @@ pki.RDNAttributesAsArray = function(rdn, md) {
       obj.type = asn1.derToOid(attr.value[0].value);
       obj.value = attr.value[1].value;
       obj.valueTagClass = attr.value[1].type;
+      if(obj.valueTagClass === asn1.Type.UTF8) {
+        obj.value = forge.util.decodeUtf8(obj.value);
+      }
       // if the OID is known, get its name and short name
       if(obj.type in oids) {
         obj.name = oids[obj.type];


dlongley:
`obj.value` is a binary-encoded string. Forge was written before TypedArrays and Buffers were available in JavaScript, so that is how it represents binary data. In the future (version 0.7.x+), it will be a buffer object to make this more clear.

In other words, you should treat `obj.value` as if it was a buffer of bytes, not a string. So we don't want to do this encoding conversion here -- it should be done outside of this code by using the same thing that's in this patch. Essentially check for the encoding type and make the appropriate call to convert the byte buffer (right now, just a binary-encoded string) to a regular JavaScript string. Similarly, when setting the value, it should be encoded as UTF-8 if the that's the desired encoding. The confusion is entirely understood -- and a future version of the API will rectify this by forcing you to use a buffer (or by allowing you to pass a string and the desired encoding -- which will convert it to a buffer for you; no longer will a string be used internally at all for `obj.value`).

pk13nv:
OK, I'll try fix this issue in pkcs7.js on line
issuer: forge.pki.RDNAttributesAsArray(capture.issuer)

... or somewhere else - it takes some time to find better place.

Thanks, PK13NV

smuda:
I just stumbled upon this behavior as well when parsing a certificate. I did a fix, very similar to PK13NV, including unit tests (at https://github.com/AyoyAB/forge). Even though dlongley said we should treat attribute.value as a buffer of bytes, could we not add a new property which is the "decoded" string? 

Something like 'valueString'?

dlongley:
@smuda,

In forge 0.7.x, the goal is to make `value` a buffer object that you can call `toString` on.

smuda:
@dlongley That sounds good to me!

dlongley:
Closing as the proper solution is to use a buffer object once we implement this in 0.8.x.

PR #411, opened 2016-07-09
Bug fix invalid crl distribution points

nixgadget:
- This is a bug fix to wrap the CRL Dist points in CHOICE as given in [rfc5280](https://tools.ietf.org/html/rfc5280#section-4.2.1.13)
- Otherwise it would fail cert verification: `openssl verify -CAfile <ca> <cert>`


nixgadget:
The tests pass on my local 6.2 branch. If you can may be retry the tests.

Issue #420, opened 2016-08-03
possible caStore.addCertificate bug for CA certificates with the same subject


zachmart:
Hi,

I was looking at the code for the caStore in the x509.js file and it seems there might be a bug that occurs when two certificates with the same subject are added. Although this is probably rare, it seems that the intention of the code is to change the caStore.certs mapping from a subject hash-> cert map to a subject hash -> array of certs map. Here is some code that illustrates the issue:

```js
var forge = require("node-forge");

var keyPair1 = forge.pki.rsa.generateKeyPair(1024);
var keyPair2 = forge.pki.rsa.generateKeyPair(1024);

var ca1 = forge.pki.createCertificate();
var ca2 = forge.pki.createCertificate();

ca1.publicKey = keyPair1.publicKey;
ca2.publicKey = keyPair2.publicKey;

ca1.serialNumber = "01";
ca2.serialNumber = "02";

ca1.validity.notBefore = new Date();
ca2.validity.notBefore = new Date();

ca1.validity.notAfter = new Date();
ca1.validity.notAfter.setFullYear(ca1.validity.notBefore.getFullYear() + 1);
ca2.validity.notAfter = new Date();
ca2.validity.notAfter.setFullYear(ca2.validity.notBefore.getFullYear() + 2);

var attrs = [
  {
    name: 'commonName',
    value: 'Cert Authority'
  }
];

// note that both CAs have the same subject distinguished name
ca1.setSubject(attrs);
ca2.setSubject(attrs);

ca1.setExtensions([
  {
    name: 'basicConstraints',
    cA: true,
    pathLenConstraint: 4
  }, {
    name: 'keyUsage',
    keyCertSign: true,
    digitalSignature: true,
    nonRepudiation: true,
    keyEncipherment: true,
    dataEncipherment: true
  }
]);

ca2.setExtensions([
  {
    name: 'basicConstraints',
    cA: true,
    pathLenConstraint: 2
  }, {
    name: 'keyUsage',
    keyCertSign: true,
    digitalSignature: true
  }
]);

ca1.sign(keyPair1.privateKey);
ca2.sign(keyPair2.privateKey);

var caStore = forge.pki.createCaStore();
caStore.addCertificate(ca1);
caStore.addCertificate(ca2);

console.log("caStore has ca1:", caStore.hasCertificate(ca1)); 
console.log("caStore has ca2:", caStore.hasCertificate(ca2)); 

// outputs:
// caStore has ca1: true
[...]

PR #421, opened 2016-08-03
CA store updates

zachmart:
I wrote these updates to the caStore to achieve several goals:

- Fix the bug I already outlined in issue #420 
- Make the methods of the caStore more consistent in the types of certificates they accept. Specifically, the addCertificate method accepts a pki.certficate object or a PEM string, but the hasCertificate method only accepts a pki.certificate. My pull request, gives all methods of caStore that accept a certificate as a parameter the ability to receive either a pki.certificate or a PEM string, including the added methods (see below).
- Add a listAllCertificates method, which allows a caStore to be serialized for storage in a database or for sending across a network and then restored at a later time.
- Add a removeCertificate method
- Add testing for this new functionality in the form of the tests/nodejs-castore.js file

### DIFF
--- a/js/x509.js
+++ b/js/x509.js
@@ -2828,6 +2828,10 @@ pki.createCaStore = function(certs) {
    */
   caStore.removeCertificate = function(cert) {
     var result;
+    // convert from pem if necessary
+    if(typeof cert === 'string') {
+      cert = forge.pki.certificateFromPem(cert);
+    }
     assureSubjectHasHash(cert.subject);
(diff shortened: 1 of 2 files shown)


dlongley:
Landed as f785e40b4e24d8ed06ee81bfd0d1b41bf33a4ab3 and 10b2edea875f43042993ad17a1739b51d14a6c14 and 3cdaf6c76ca72bd70d9b642c038fb46355857456.

Thanks @zachmart!

PR #434, opened 2016-10-10
Add couple of missing oids

mgcrea:
Just a couple of missing oids for better identification when parsing exts.


dlongley:
Thanks @mgcrea!

PR #357, opened 2016-02-09
Switch to CommonJS

ysangkok:
This removes the boilerplate. All tests in `/nodejs/test` pass, including the browser test.

PR #445, opened 2016-11-30
Fix Node module loading

mdebruijne:
This patch fixes module loading on Node.js.

### DIFF
--- a/js/forge.js
+++ b/js/forge.js
@@ -38,7 +38,9 @@ var defineFunc = function(require, module) {
     }
     forge.defined[name] = true;
     for(var i = 0; i < mods.length; ++i) {
-      mods[i](forge);
+      if(typeof mods[i] === 'function') {
+        mods[i](forge);
+      }
     }
     return forge;
   };


davidlehn:
What are the non functions that leak through to that point?
Looks like this change won't break other usage.  Anyone know if there are edge cases to be aware of?

mdebruijne:
Considering the plans in #357 I haven't spent time in analyzing/optimizing further. This fix eliminates the unexpected throw and therefore prevents crashes in applications/libraries that depend on forge. In my case Google Cloud Node.js Client, Google APIs Node.js Client and Jest.

dlongley:
I think this is fine to merge. I'm only concerned that it's covering up a problem or not fully solving it in those environments where it seems to make a difference. That being said, we're overhauling the module loading system in a new version, so I think this is a perfectly fine stop gap measure if it helps people. We should merge on those grounds, IMO.

timpler:
I've lost 2 hours looking for this. (Happened when I was trying to `import 'parse-server'` when testing with `jest`

mdebruijne:
@davidlehn , @dlongley 

Has consensus been reached that this pull request can be merged?

The current version of Forge is causing crashes in many places with corresponding upstream issue reports. 

Please let me know if I can do anything to help moving this forward. Your support in merging this pull request and releasing an update will be very much appreciated by anyone who (indirectly) depend on Forge.

davidlehn:
I merged this since it hopefully won't cause other problems.  But I was unable to reproduce the issue with basic require tests.  Can someone please post a test case is that fails?  Also since I can't test the issue I would still like to know an answer my previous question about what non-function data is that is leaking through.  If someone could console.log it that would be nice.  It would be better to fix the cause of this problem rather than hack around it.

davidlehn:
Released as 0.6.46.

mdebruijne:
Thanks!

PR #446, opened 2016-12-02
Fix issue with generation of bundled and minified forge file on Windows.

bharat-srinivasan:
* Use `r.js.cmd` to generate optimized resources when running on Windows.
* Create new file `optimize.js` which determines the OS platform and runs the correct command to perform optimization.

### DIFF
--- a/optimize.js
+++ b/optimize.js
@@ -1,0 +1,20 @@
+/**
+ * Prepares bundle for Node Forge Library based on OS platform.
+ */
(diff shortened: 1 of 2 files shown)


dlongley:
Once the changes from #456 have been published, this issue should be revisited to see if it has been addressed.

davidlehn:
0.7.0 was released which uses webpack by default.  This should address this issue.

PR #460, opened 2016-12-19
Add Simple Certificate Enrollment Protocol (SCEP) OIDs

sbruton:
These OIDs are necessary for attributes within PKCS#7 messages built for SCEP requests and responses.


sbruton:
needs some more work, missing OIDs was not the extent of the ASN1 gaps for SCEP support

Issue #468, opened 2017-01-12
Allow selective use of md algorithms.


davidlehn:
As mentioned in https://github.com/digitalbazaar/forge/issues/467, it should be possible to combine files without requiring all of the algorithms from md.js.

0.6.x was a bit more relaxed in what files got used but code conversion for 0.7.x made it more aggressive to just use all algorithms via the generic md.js.

One solution is to have md.js only setup the simple data structures and then individual algorithm files (ie, sha256.js) can be required by higher level code as needed.  As a default, index.js (or some meta md file) could require all known algorithms.

davidlehn:
@alexlamsl: I think this addresses your concern from #467.  You should be able to just bundle pbkdf2.js and sha256.js and the dependencies will pull in the other files.  An example was added to the webpack file to do this.  Please reopen if this doesn't fit your use case. Thanks.

alexlamsl:
@davidlehn looks great, thanks!

PR #471, opened 2017-01-16
Require minimal digest algorithm dependencies.

davidlehn:
Trying to address concerns of https://github.com/digitalbazaar/forge/issues/468


davidlehn:
Note this could also be done to `pbe.js`. There is some code that can select different algorithms. It could change to not include all by default and throw an error if an algorithm is not found. Does it make sense in that file?

Is the `md.all.js` name ok?

dlongley:
It's fine to throw errors for people that are doing custom builds and trying to minimize included code (same goes for pbe). The name `md.all.js` is fine too. LGTM.

dlongley:
This is a related note that a significant amount of work has already been put into a new MessageDigest API and consolidation of common code in the 0.8.x branch. I just want to make sure we don't duplicate efforts at all (I don't think we are yet, just a warning):

https://github.com/digitalbazaar/forge/blob/0.8.x/js/md.js

davidlehn:
`pbe.js` updated now too. Will leave other use cases for specific requests or 0.8.x.

PR #474, opened 2017-01-23
Fix parsing content of PKCS#7-message if it is DER-data

anthrax63:
If i use DER-data as a "content" of PKCS#7 message then after i parse it from PEM-format i get the ByteString with zero length instead of valid DER-bytes. This commit fixes that problem.

### DIFF
--- a/lib/pkcs7.js
+++ b/lib/pkcs7.js
@@ -1183,6 +1183,8 @@ function _fromAsn1(msg, obj, validator) {
         }
         content += capture.content[i].value;
       }
+    } else if (typeof capture.content === 'object') {
+      content = asn1.toDer(capture.content).getBytes();
     } else {
       content = capture.content;
     }
--- a/tests/unit/pkcs7.js
+++ b/tests/unit/pkcs7.js
@@ -1,6 +1,7 @@
 var ASSERT = require('assert');
 var forge = require('../../lib/forge');
 var PKCS7 = require('../../lib/pkcs7');
+var ASN1 = require('../../lib/asn1');
 var PKI = require('../../lib/pki');
 var AES = require('../../lib/aes');
 var DES = require('../../lib/des');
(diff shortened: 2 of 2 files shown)


lib/pkcs7.js
```diff
@@ -342,16 +342,20 @@ p7.createSignedData = function() {
         if('content' in msg) {
           var content;
           if(msg.content instanceof forge.util.ByteBuffer) {
-            content = msg.content.bytes();
+            content = asn1.create(asn1.Class.UNIVERSAL, asn1.Type.OCTETSTRING, false,
```
dlongley: I believe this would break the design here. The format for the `content` of PKCS#7 messages is dependent on the `contentType`. The current design allows for people to specify a `ByteBuffer` (or binary string of bytes) with DER-encoded content -- that expresses whatever ASN.1 format they desire. This would force the content to be an `asn1.Type.OCTETSTRING`, which is undesirable.

dlongley:
I may have misread these changes ... looking back over it again.

lib/pkcs7.js
```diff
@@ -342,16 +342,20 @@ p7.createSignedData = function() {
...
+              msg.content.bytes());
           } else if(typeof msg.content === 'string') {
-            content = forge.util.encodeUtf8(msg.content);
+            content = asn1.create(asn1.Class.UNIVERSAL, asn1.Type.OCTETSTRING, false,
+              forge.util.encodeUtf8(msg.content));
+          } else if (typeof msg.content === 'object') {
+            content = asn1.create(asn1.Class.UNIVERSAL, asn1.Type.OCTETSTRING, true,
+              [msg.content]);
```
dlongley: So, looking back at this code, this section of code only executes as a helper to automatically generate `contentInfo` in a common case. I think there was a presumption that the `contentInfo` would be generated externally unless `msg.content` is set to a byte buffer or a string. However, the conditional seems to allow this code to execute if `msg.content` is an `object` ... and I'm not sure why yet. I'll need to see if there is well defined behavior in this case in the documentation, but it seems to me that `msg.content` as an `object` could imply a number of different things and we can't simply pick to encode it like this.

In other words, if your `msg.content` is a string or a byte buffer, we have a well defined way in which it will be automatically wrapper. If you need to do something else, you should generate your own `contentInfo`. If a simple tweaking to this code better handles some other common cases that's fine, but we need to say explicitly what's going to happen if you pass in `msg.content` as an object -- and it's not yet clear to me that what's going on here is the "right" thing (or the highest utility thing) to do.

dlongley: I'm also not yet sure how my above comment meshes with parsing PKCS#7 messages, so I'll have to look back at that.

tests/unit/pkcs7.js
```diff
@@ -518,6 +519,23 @@ var support = require('./support');
...
+      p7.content = UTIL.createBuffer(ASN1.toDer(ASN1.create(ASN1.Class.UNIVERSAL, ASN1.Type.OCTETSTRING, false, 'Just a little test')));
+      p7.encrypt();
+
+      var pem = PKCS7.messageToPem(p7);
+
+      // convert back from PEM to new PKCS#7 object, decrypt, and test
+      p7 = PKCS7.messageFromPem(pem);
+      p7.decrypt(p7.recipients[0], PKI.privateKeyFromPem(_pem.privateKey));
+      var asn1 = ASN1.fromDer(p7.content);
+      ASSERT.equal(asn1.value, 'Just a little test');
```
dlongley: This test case looks good from what I can tell. Will look more closely when I get a chance.

lib/pkcs7.js
dlongley: Nevermind -- I see that the OCTETSTRING below was, in effect, moved up here (this is not a new one that wraps it).

GitStorageOne:
No activity since 2017?

PR #476, opened 2017-01-25
Certificate parsing fixes.

davidlehn:
- Fixes possible issues with certificate parsing.
- Eliminates many parsing failures. Often due to parsing BIT STREAMs as possible ASN.1 data.
- Improve many checks to protect against getting out of sync with byte stream.
- Add support for "raw" data to ensure BIT STREAM data stays constant while capturing or encoding when it may have mistakenly been interpreted as ASN.1 data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Forge ChangeLog
 ### Fixed
 
 - Fix test looping bugs so all tests are run.
+- Improved ASN.1 parsing. Many failure cases eliminated. More sanity checks.
+  Better behavior in default mode of parsing BIT STRINGs. Better handling of
+  parsed BIT STRINGs in `toDer()`. More tests.
 
 ### Changed
 
@@ -20,6 +23,9 @@ Forge ChangeLog
   [forge-dist](https://github.com/digitalbazaar/forge-dist).
 - **BREAKING**: Require minimal digest algorithm dependencies from individual
   modules.
+- Enforce currently supported bit param values for byte buffer access. May be
+  **BREAKING** for code that depended on unspecified and/or incorrect behavior.
+- Improve `asn1.prettyPrint()` BIT STRING display.
 
 ### Added
 
@@ -41,6 +47,8 @@ Forge ChangeLog
 - Add rsa.generateKeyPair async and usePureJavaScript tests.
 - Add .editorconfig support.
 - Add `md.all.js` which includes all digest algorithms.
+- Add asn1 `equals()` and `copy()`.
+- Add asn1 `validate()` capture options for BIT STRING contents and value.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Forge ChangeLog
 - Improved ASN.1 parsing. Many failure cases eliminated. More sanity checks.
   Better behavior in default mode of parsing BIT STRINGs. Better handling of
   parsed BIT STRINGs in `toDer()`. More tests.
+- Improve X.509 BIT STRING handling by using new capture modes.
 
 ### Changed
 
--- a/lib/asn1.js
+++ b/lib/asn1.js
@@ -323,7 +323,7 @@ asn1.equals = function(obj1, obj2, options) {
  *
  * @return the length of the BER-encoded ASN.1 value or undefined.
  */
-var _getValueLength = asn1.getBerValueLength = function(b) {
+asn1.getBerValueLength = function(b) {
   // TODO: move this function and related DER/BER functions to a der.js
   // file; better abstract ASN.1 away from der/ber.
   var b2 = b.getByte();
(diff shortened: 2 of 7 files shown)


davidlehn:
Note these fixes also cleanly apply to 0.6.x and a branch is available:
https://github.com/digitalbazaar/forge/tree/cert-fixes-0.6.x

lib/asn1.js
```diff
@@ -248,141 +254,277 @@ var _getValueLength = asn1.getBerValueLength = function(b) {
 };
 
 /**
+ * Check the byte buffer has enough bytes. Throws an Error if not.
```
dlongley: "Check the byte" => "Check if the byte"

lib/asn1.js
```diff
@@ -206,13 +208,17 @@ asn1.create = function(tagClass, type, constructed, value) {
...
-  return {
+  var obj = {
     tagClass: tagClass,
     type: type,
     constructed: constructed,
     composed: constructed || forge.util.isArray(value),
     value: value
   };
+  if(options && 'raw' in options) {
+    obj.raw = options.raw;
```
dlongley: I'm not a huge fan of the name `raw`. I don't feel like it's descriptive enough for what's going on, but since I don't have a better name, we can go with it. It's not difficult to alias later and deprecate.

lib/asn1.js
```diff
@@ -248,141 +254,277 @@ var _getValueLength = asn1.getBerValueLength = function(b) {
...
  *
  * @param bytes the byte buffer to parse from.
- * @param strict true to be strict when checking value lengths, false to
+ * @param [strict] true to be strict when checking value lengths, false to
  *          allow truncated values (default: true).
+ * @param [options] object with options or boolean strict flag
+ *          [strict] true to be strict when checking value lengths, false to
+ *            allow truncated values (default: true).
+ *          [decodeBinaryStrings] true to attempt to decode the content of
+ *            BIT STRINGs (not OCTET STRINGs) in strict mode. Note that without
```
dlongley: Since this only applies to `BIT STRINGS`, let's call the option `decodeBitStrings`. I think our hope is to never support it with `OCTET STRINGS` because the better implementation would be to use a schema if you want this sort of thing.

dlongley: We should also mention that this flag is a hack that should only be used when absolutely necessary -- and that the schema support mentioned will be a future feature for a robust way to deal with the issue.

lib/asn1.js
```diff
@@ -248,141 +254,277 @@ var _getValueLength = asn1.getBerValueLength = function(b) {
...
   // wrap in buffer if needed
   if(typeof bytes === 'string') {
     bytes = forge.util.createBuffer(bytes);
   }
 
+  return _fromDer(bytes, bytes.length(), 0, options);
+}
+
+/**
+ * Internal functino to parse an asn1 object from a byte buffer in DER format.
```
dlongley: Typo

lib/asn1.js
```diff
@@ -916,6 +1062,9 @@ asn1.validate = function(obj, v, capture, errors) {
         if(v.captureAsn1) {
           capture[v.captureAsn1] = obj;
         }
+        if(v.captureRaw && 'raw' in obj) {
```
dlongley: I'm wondering if `captureDer` would be a better name. It wouldn't work map cleanly with BER-formatted data, unfortunately.

lib/asn1.js
```diff
@@ -997,12 +1146,12 @@ asn1.prettyPrint = function(obj, level, indentation) {
...
       break;
-    case asn1.Type.BITSTRING:
-      rval += ' (Bit string)';
-      break;
     case asn1.Type.INTEGER:
       rval += ' (Integer)';
       break;
+    case asn1.Type.BITSTRING:
+      rval += ' (Bit string)';
+      break;
```
dlongley: Why move this?

davidlehn: Moved to be in the same order as the type ids.

lib/asn1.js
davidlehn: Understood. Bits? Bytes? Octets? Data?

lib/asn1.js
davidlehn: I initially thought it might control OCTET STRING behavior too, but yes, more direct name now and alternative solutions later sounds good.

davidlehn:
Force pushed updates and fixes.
@dlongley: Please review again.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ Forge ChangeLog
 ### Fixed
 
 - Fix test looping bugs so all tests are run.
+- Improved ASN.1 parsing. Many failure cases eliminated. More sanity checks.
+  Better behavior in default mode of parsing BIT STRINGs. Beter handling of
```
dlongley: `Beter` => `Better`

lib/asn1.js
```diff
@@ -206,13 +209,103 @@ asn1.create = function(tagClass, type, constructed, value) {
...
+  if(forge.util.isArray(obj)) {
+    copy = [];
+    for(var i = 0; i < obj.length; ++i) {
+      copy.push(asn1.copy(obj[i], options));
+    }
+    return copy;
+  }
+
+  if(typeof obj === 'string') {
+    return String(obj);
```
dlongley: `String()` cast unnecessary

lib/asn1.js
```diff
@@ -206,13 +209,103 @@ asn1.create = function(tagClass, type, constructed, value) {
...
+
+  copy = {
+    tagClass: obj.tagClass,
+    type: obj.type,
+    constructed: obj.constructed,
+    composed: obj.composed,
+    value: asn1.copy(obj.value, options)
+  };
+  if(options && !options.excludeBitStringContents) {
+    copy.bitStringContents = String(obj.bitStringContents);
```
dlongley: `String()` cast unnecessary. Add a `TODO: copy byte buffer if its a buffer not a string` for the future.

davidlehn:
@dlongley Please review latest commits. Thanks.

PR #493, opened 2017-03-10
set correct digest size for hashes based on SHA-512

linuxwolf:
Without this change, PSS (using something other than sha1 / sha256 / sha512) could be incorrectly calculated, thereby throwing a `index out of range` error.

### DIFF
--- a/lib/sha512.js
+++ b/lib/sha512.js
@@ -79,12 +79,26 @@ sha512.create = function(algorithm) {
     _w[wi] = new Array(2);
   }
 
+  // determine digest length by algorithm name (default)
+  var digestLength = 64;
+  switch (algorithm) {
+    case 'SHA-384':
+      digestLength = 48;
+      break;
+    case 'SHA-512/256':
+      digestLength = 32;
+      break;
+    case 'SHA-512/224':
+      digestLength = 28;
+      break;
+  }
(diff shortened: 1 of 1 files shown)


davidlehn:
It would be better to set the correct digestLength in `sha512.create` so other callers to that function will get the right length.  Probably should leave it unset when the `md` structure is setup then set it based on the normalized `md.algorithm` value.

linuxwolf:
I can make that change this weekend for review

linuxwolf:
@davidlehn  I changed the PR to do all the work in `sha512.create`, based on the algorithm name.  I think this is closer to what is preferred.

Also, it's not obvious the right way to add tests for this.  The easiest is to tweak the `sha512` to check the value of `.digestLength`, but it might be more comprehensive to add another set of PSS tests ... somehow.  If you can provide some guidance, I'd be happy to take a first pass at them.

msporny:
@davidlehn ping, pushing this to the top of your stack. Can you please get back to @linuxwolf in the next day or two about this PR?

davidlehn:
I added really basic digestLength tests.  Feel free to add pss tests that depend on correct digest lengths.  I'm not familiar enough with that part of the code to suggest how to do so.

PR #501, opened 2017-04-18
Fix README header formatting

BigBlueHat:
The Markdown (at least on GitHub) was not catching headings
just below the anchor tags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,7 @@ API
 ---
 
 <a name="options" />
+
 ### Options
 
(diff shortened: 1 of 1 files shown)


davidlehn:
Thanks.  I'm guessing they recently updated their markdown libs and broke this pattern.  It was certainly working last time I looked.

BigBlueHat:
Yep! Related to standards of all things!! :grin:
https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown

PR #386, opened 2016-04-25
Skip validity period check

RoomUnitBDRThermea:
Made sure the pull request is now based on the latest 0.6.x release of Forge.

Issue #222, opened 2015-02-27
enable pki.verifyCertificateChain() to test with an arbitrary date/time (for S/MIME email)


rhashimoto:
This is a feature request.

`pki.verifyCertificateChain()` checks that all chain certificates are valid at the current date/time. It would be nice if an optional `Date` could be provided for this test. An example of why this might be useful is verifying S/MIME signed email, where all certificates were valid at the time the message was sent but one or more have since expired. It doesn't appear possible to use this function for this purpose.

Alternatively, if the date/time check is moved to be the last one before calling the `verify` callback, then no change in the function arguments is needed as the callback can override the error knowing that all other checks have passed. This might break existing code as the `verify` callback could _not_ be used to override the check that is currently last (`keyUsageExt`), though this seems unlikely.

A third possible solution would be to always apply all the tests and pass a collection of errors to the `verify` callback. Unfortunately this is more likely to break existing code.

dlongley:
This is similar to #188. We agree there's value in a feature like this and we'd like to support it -- we just need some time to sit down and figure out the right way forward. Ideally, we wouldn't break backward compatibility. If we must, then the feature can be part of **0.7.x** which has some other breaking API changes.

dlongley:
Related to #195 as well.

davidlehn:
A variation of this was added in https://github.com/digitalbazaar/forge/pull/635.

PR #502, opened 2017-04-20
Add optional date parameter to pki.verifyCertificateChain

oyvped:
Add an optional date to pki.verifyCertificateChain to check for validity on a specific date. This was included in the old 0.7.x branch which is now called 0.8.x.

### DIFF
--- a/lib/x509.js
+++ b/lib/x509.js
@@ -2899,6 +2899,7 @@ pki.certificateError = {
  * @param chain the certificate chain to verify, with the root or highest
  *          authority at the end (an array of certificates).
  * @param verify called for every certificate in the chain.
+ * @param date optional date to check for validity. Defaults to current date.
  *
  * The verify callback has the following signature:
  *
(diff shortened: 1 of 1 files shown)


davidlehn:
See also: https://github.com/digitalbazaar/forge/pull/386/files

davidlehn:
I think we wanted to use some sort of options param to be more flexible with possible checks.  It probably makes sense to support some common things like date but may also make sense to also have a generic check function option that gets called for every cert so user code can do what checks it wants.  (Though maybe just checking the cert chain yourself is easy enough?)

davidlehn:
Also https://github.com/digitalbazaar/forge/issues/222

hardersen:
Right, what if we add an optional options parameter to `verifyCeriticateChain` with an optional `verifyCertificateDate(certificate)` function that if specified overrides the current date check?

It's hard to use Forge to verify a historic signature if one can't specify the signing date. It's a common use case - for instance OpenSSL has `attime` in the `verify` command.

davidlehn:
A variation of this was added in https://github.com/digitalbazaar/forge/pull/635.

Issue #520, opened 2017-07-06
Some serious Fortuna implementation issues


nizkp:
First of all, I have to say that this library has been very useful for me, and I appreciate the great work that you have done to implement all the algorithms. I used it to provide a JS only implementation of some subset of WebCrypto that we use in our company, for a code that was ported to run in a  React-Native environment. Since WebCrypto or nodejs crypto APIs are not available, I was concerned about the crypto random number generator, wanted to make sure it was properly seeded, etc.

So I did a quick review of the PRNG implementation (Fortuna) and I found some deviations from the paper (https://www.schneier.com/academic/paperfiles/fortuna.pdf) which I don't understand (as far as I know these are mistakes). I think some are quite obvious, which makes me think that no serious review/audit of the PRNG implementation has been done. I'm not saying that the implemented PRNG is insecure, but more that the intended algorithm (Fortuna) is not implemented properly, and its security relies **A LOT** on `crypto.getRandomValues` or the equivalent NodeJS API.

**My concerns:**

1. A very concrete one: in function `_seed `in prng.js, there is the comment:
`  // digest the entropy of other pools whose index k meet the`
`  // condition '2^k mod n == 0' where n is the number of reseeds`
  Please, check the paper, it's actually the other way around: `'n mod 2^k == 0'`. It's not just the comment, the code is wrong. Note that this also works for bucket k = 0, since n mod 2^0 is always 0, so bucket 0 is always used.

2. If pool 0 does not have enough collected data, or not enough time since last reseed, there should not be a reseed. But in your implementation you just force a   reseed for every call to `getBytesSync` and if pool 0 does not have the minimum entropy you just get the remaining from the 'seed file', which I believe it's not the intended usage for this. AFAIK the seed file should only be used as the initial seed. In fact, IMHO probably it would be better if this concept was replaced by just an 'initial seed', since there is no actual file here (are you writing to it?). But instead you use this "seed file" as a magical stream of random data, which seems a bit like cheating.

3. According to the paper reseed should be done like `newkey = sha256(sha256(oldkey ||| entropy))`, where entropy is the accumulated pools/buckeys entropy. This is good because the security
[...]

PR #521, opened 2017-07-07
Fix Fortuna implementation bugs.

dlongley:
Addresses #520.

This isn't fully done yet, but it addresses:

1. Fixing the incorrect (reversed) condition for selecting pools to use entropy from when reseeding.
2. Includes the old key bytes when reseeding.

Some other concerns were raised in #520 that may need to be addressed -- or they may be ok, more research is needed.


dlongley:
Going to merge this for now, we can add more PRs to address other issues or add further testing.

PR #523, opened 2017-07-12
Fix issue where encrypted data was not decryptable

diffidentDude:
We are encrypting on the front end using RSA-OAEP. 

    const encrypted = key.encrypt(data, 'RSA-OAEP', {
      md: Forge.md.sha256.create(),
      mgf1: {
        md: Forge.md.sha256.create()
      }
    });

We found that we were not able to decrypt with either bouncy castle or the HSM. 

After debugging with forge and bouncy castle side by side we found this was the only difference between the libraries.

### DIFF
--- a/lib/pkcs1.js
+++ b/lib/pkcs1.js
@@ -142,7 +142,7 @@ pkcs1.encode_rsa_oaep = function(key, message, options) {
   var maskedSeed = forge.util.xorBytes(seed, seedMask, seed.length);
 
   // return encoded message
-  return '\x00' + maskedSeed + maskedDB;
+  return maskedSeed + maskedDB;
 };
 
 /**


dlongley:
Prepending the leading `0x00` byte is correct according to [PKCS#1 v2.1](https://www.ietf.org/rfc/rfc3447.txt). This is a change from 2.0, but it should be backwards compatible. So the patch in this PR would not be correct -- there must be some other issue with forge or on the bouncy castle side.

https://crypto.stackexchange.com/questions/40032/why-did-oaep-change-from-pkcs1-v2-0-and-v2-1

dlongley:
The latest, [PKCS#1 2.2](https://tools.ietf.org/html/rfc8017#section-7.1.1), also prepends a `0x00` byte to the encoded message.

diffidentDude:
Cheers for the feedback, we'll do some more investigation and get back to you.

PR #544, opened 2017-11-11
asn1.equals: Compare all array elements, not just the first

andersk:
Discovered by the Closure Compiler:

    $ closure-compiler --jscomp_warning=lintChecks dist/forge.all.js
    […]
    dist/forge.all.js:3551: WARNING - unreachable code
        for(var i = 0; i < obj1.length; ++i) {
                                        ^^^


davidlehn:
Thanks!

PR #565, opened 2018-02-22
[pkcs7] Import signedData message with no certificates

walac:
RFC 2315, section 9.1 explicitly says that the certificates attribute is
optional.

### DIFF
--- a/lib/pkcs7.js
+++ b/lib/pkcs7.js
@@ -142,9 +142,11 @@ p7.createSignedData = function() {
       msg.contentInfo = null;
       msg.signerInfos = [];
 
-      var certs = msg.rawCapture.certificates.value;
-      for(var i = 0; i < certs.length; ++i) {
-        msg.certificates.push(forge.pki.certificateFromAsn1(certs[i]));
+      if(msg.rawCapture.certificates) {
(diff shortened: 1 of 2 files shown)


lib/pkcs7.js
```diff
@@ -142,9 +142,11 @@ p7.createSignedData = function() {
       msg.contentInfo = null;
       msg.signerInfos = [];
 
-      var certs = msg.rawCapture.certificates.value;
-      for(var i = 0; i < certs.length; ++i) {
-        msg.certificates.push(forge.pki.certificateFromAsn1(certs[i]));
+      if (msg.rawCapture.certificates) {
```
dlongley: Minor style nit (no space after `if`).

walac: Fixed :)

davidlehn:
I just decoded the pkcs7 test data, are you ok with the JSON blob with AWS instanceId, accountId, imageId, privateIp, etc going in the forge repo?  It's already here on github now, but it might be better if it was more anonymous data.

walac:
This instance is gone a long time ago. And none of this information is private.

davidlehn:
Thanks!

Issue #563, opened 2018-02-15
Potential catastrophic backtracking regex vulnerability


msporny:
From @davisjam via email::

Forge has a regex vulnerable to catastrophic backtracking, which may lead to poor performance (if used client-side) or a REDOS vulnerability (if used server-side, esp. on attacker-controlled input).

I am not sure how likely this is in your case, but given the popularity of your module (6.8M downloads/month!) I thought I would let you know.

Regex:

lib/form.js
```
var _regex = /(.*?)\[(.*?)\]/g;
```
(introduced in 2011, 6d4e32e4)

Here is a sample attack string:
```
{"pumpPairs":[{"pump":"[","prefix":"a"}],"suffix":"["}
```

i.e. "prefix" + "pump a bunch of times" + "suffix".

The catastrophic backtracking follows a power law: f(x) ~= a * x^2.58, where x is the number of pumps. On my machine, about 3K pumps (about 3K chars) blocks the node event loop for 10 seconds.

Fix advice
If you want to fix it, you can refactor the regex (the problem lies in the adjacent overlapping quantified classes `.* ... .*`) . You could also replace it with custom parsing. If legitimate input strings should be no longer than, say, a few hundred characters you could also simply trim the input.

Let me know if you have any questions.

If you are not familiar with REDOS, here are some good starting points for information:
- https://docs.microsoft.com/en-us/dotnet/standard/base-types/backtracking-in-regular-expressions
- https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices#Backtracking
- https://snyk.io/blog/redos-and-catastrophic-backtracking

davidlehn:
The issue title is perhaps a bit too dramatic.  The `form.js` part of this library is ancient, unrelated to the crypto parts, probably not used by anyone including the authors, and likely to be purged at some point.

murgatroid99:
I hope I didn't miss something, but I think this problem can be solved without changing the functionality by changing the regex to `/([^[]*?)\[(.*?)\]/g`. The use of the lazy star quantifier in the first capturing group in the original regex indicates that for any successful match, the first capturing group should never contain a `[`. So, if we enforce that explicitly, we should be able to avoid this catastrophic backtracking. Does that look right?

davisjam:
The pattern you propose is safe because it removes the ambiguity of the current pattern. Once a [ is seen the engine can't backtrack back to the first group.

[...]

PR #570, opened 2018-03-07
Fix catastrophic backtracking

davisjam:
One regex in this project was vulnerable
to catastrophic backtracking.

I have replaced it with a safe regex that matches
a nearly-equivalent language.
I believe this language is still correct for the expected input.

This is a fix for #563.


davidlehn:
Thanks!

PR #534, opened 2017-08-24
TLS 1.2 implementation

parthenon:
This pull request consists of our implementation of TLS1.2 support, which we added for one of our projects. It supports two new cipher suites: TLS_RSA_WITH_AES_128_CBC_SHA256 and TLS_RSA_WITH_AES_256_CBC_SHA256. Tested against the unit test suite and openssl s_server with TLS 1.1 and TLS 1.2 forced, as well as examples/tls.js and examples/imap.js.

PR #581, opened 2018-05-05
TLS 1.2 client implementation 

githoniel:
base on @parthenon's [PR](https://github.com/digitalbazaar/forge/pull/534) 

Tested with node offical tls server/Go offical tls server

- fix two syntax error
- `handleCertificateRequest` add support for SignatureAndHashAlgorithm
- `getClientSignature`  support TLS 1.2 SignatureAlgorithm using SHA256-RSA
- `createCertificateVerify` add add support for SignatureAndHashAlgorithm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .cdtproject
 .classpath
 .cproject
+.nyc_output
 .project
 .settings
 TAGS
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: node_js
 node_js:
   - "4"
   - "6"
+  - "8"
+  - "10"
   - "node"
 sudo: false
 install: npm install
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,36 @@
 Forge ChangeLog
 ===============
 
+## 0.7.5 - 2018-03-30
+
+### Fixed
+- Remove use of `const`.
+
+## 0.7.4 - 2018-03-07
+
+### Fixed
+- Potential regex denial of service in form.js.
+
+### Added
+- Support for ED25519.
+- Support for baseN/base58.
+
+## 0.7.3 - 2018-03-05
+
+- Re-publish with npm 5.6.0 due to file timestamp issues.
+
+## 0.7.2 - 2018-02-27
+
+### Added
+- Support verification of SHA-384 certificates.
+- `1.2.840.10040.4.3'`/`dsa-with-sha1` OID.
+
+### Fixed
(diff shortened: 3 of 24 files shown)


davidlehn:
Thanks for the updated patch.  As mentioned in the other PR, there was some previous TLS 1.2 work in the 0.8 branch too.  It would be good to see if any of that can be merged in as well (while dealing with the general unmerged API changes).  But maybe partial support would be ok for now.  Any chance of resolving the conflicts with master?

Did you have any automated testing scripts for this code?

matianfu:
is there any plan to merge this pr recently?

torsina:
update maybe ?

heol-files:
Hi, is this to be merged soon? TLS 1_2 would be nice to have!
Thanks for the good work!

BurakDev:
Hi @githoniel, what about server TLS 1.2 ?

githoniel:
> Hi @githoniel, what about server TLS 1.2 ?

sorry it has been a long time. I only test client code and it works. I think it will work on  server side two because they share those code

ProgrammerIn-wonderland:
this seems to cause a handshake failure on tls 1.2 websites with the code
```js
const net = require('net')
const forge = require('./forge')
var socket = new net.Socket();

var client = forge.tls.createConnection({
  server: false,
  verify: function(connection, verified, depth, certs) {
    // skip verification for testing
    console.log('[tls] server certificate verified');
    return true;
  },
  connected: function(connection) {
    console.log('[tls] connected');
    // prepare some data to send (note that the string is interpreted as
    // 'binary' encoded, which works for HTTP which only uses ASCII, use
    // forge.util.encodeUtf8(str) otherwise
    client.prepare('GET / HTTP/1.1\r\nHost: github.com\r\n\r\n');
  },
  tlsDataReady: function(connection) {
    // encrypted data is ready to be sent to the server
    var data = connection.tlsData.getBytes();
    socket.write(data, 'binary'); // encoding should be 'binary'
  },
  dataReady: function(connection) {
    // clear data from the server is ready
    var data = connection.data.getBytes();
    console.log('[tls] data received from the server: ' + data);
  },
  closed: function() {
    console.log('[tls] disconnected');
  },
  error: function(connection, error) {
    throw error
  }
});

socket.on('connect', function() {
  console.log('[socket] connected');
  client.handshake();
});
socket.on('data', function(data) {
  client.process(data.toString('binary')); // encoding should be 'binary'
});
socket.on('end', function() {
  console.log('[socket] disconnected');
});

// connect to google.com
socket.connect(443, 'github.com');

// or connect to gmail's imap server (but don't send the HTTP header above)
//socket.connect(993, 'imap.gmail.com');
````

PR #586, opened 2018-06-01
Detect WebCrypto API in web workers via `self`

leoselig:
## Problem
The following code has been used to detect the Web Crypto API:
```javascript
function _detectSubtleCrypto(fn) {
  return (typeof window !== 'undefined' &&
    typeof window.crypto === 'object' &&
    typeof window.crypto.subtle === 'object' &&
    typeof window.crypto.subtle[fn] === 'function');
}
```

While this works fine in the browser main thread, it fails to detect the crypto API in Web Workers.

## Solution
A typical solution to write code that works in Web Workers as well as the main thread is to use the [`self` keyword](https://developer.mozilla.org/en-US/docs/Web/API/Window/self) that resolves to `window` in the main thread and the instance of `WorkerGlobalScope` in the web worker.

Since this is not supported yet in `< IE 11`, I'm introducing a `getGlobalScope()`-helper in the `util`-namespace.

As we are using `node-forge` in a web worker to avoid blocking the main thread for other operations, this sped up RSA key pair generation from ~1min in average to less than 3 seconds.

Feel free to tell me what should change to make this happen. :)


davidlehn:
Is there an easy way to test this?  Maybe some sort of modification to the legacy keygen interactive test?

leoselig:
@davidlehn I wasn't very comfortable adding something to a "legacy" setup so I took the liberty of tweaking karma setup a bit. It now allows karma-only tests. This allowed me to setup a web worker without compromising the mocha tests (that only run in Node.js).

This does not reliably test the performance improvement but hopefully serves as a starting point to ensure, i.e., compatibility with web workers (or even other browser-only APIs).

I'm aware that this might have been too big of a leap forward, so be sure that I'm appreciating whatever feedback is incoming . :)

leoselig:
I just realized, browserify is available as a bundler option.
Any particular reason to support both webpack and browserify?
As you can see by the use of the `worker-loader`, only supporting one or the other would be quite simpler.

davidlehn:
I like that approach of having karma specific tests.  I'll take a look at that test when I get some time.

I'm not sure how many people use browserify.  It wasn't much effort to keep it working so we just kept it in the test setup.  I imagine a future forge is going to go all-in on ES2017+ which is when I figured all the babel transpiling and polyfills would make browserify path more difficult.  If it's possible to keep the tests working now, that would be nice.  Not quite sure what that involves here.  Maybe easier to just disable difficult things in browserify.

leoselig:
Hey @davidlehn 
Did you manage to take a closer look already? Any way I can help to finish this?

My proposal would be to drop support for running tests based on browserify. As far as I can tell there is no immediate benefit of running tests with both bundlers and webpack seems to have taken the lead in the community. If you agree, I'd be up to adapt this PR.

yieme:
Love the idea of supporting service workers.

For what it's worth, I suspect the following may be cleaner and reduce the overhead of a function call:

```js
util.getGlobalScope = this.global // node 
   || this.window // browser 
   || this.self // service worker
```

lib/rsa.js
```diff
@@ -1735,10 +1739,10 @@ function _getMillerRabinTests(bits) {
  * @return true if detected, false if not.
  */
 function _detectSubtleCrypto(fn) {
-  return (typeof window !== 'undefined' &&
-    typeof window.crypto === 'object' &&
-    typeof window.crypto.subtle === 'object' &&
-    typeof window.crypto.subtle[fn] === 'function');
+  return (typeof util.getGlobalScope() !== 'undefined' &&
```
davidlehn: Micro optimization would be to do lift out `var globalScope = util.getGlobalScope()` above to avoid repeated calls.

davidlehn:
@leoselig Also probably should consider changing that util.getGlobalScope() function to a util.globalScope var as @yieme suggested to avoid unneeded recomputation.

leoselig:
@davidlehn Sorry to ping you once again. We just merged the upstream back into my fork. Tests are still passing, apps work as expected.
This would be a beautiful time for another merge attempt 🙏

davidlehn:
@leoselig Made some modifications here: https://github.com/digitalbazaar/forge/pull/645.
Let me know what you think.

davidlehn:
@leoselig I updated https://github.com/digitalbazaar/forge/pull/645 to keep browserify support.  I dropped your commit and my revert commits of it.  The reason to keep support is that people may be using browserify with forge.  Hard to know.  It still works so until we decide to specifically stop supporting browserify, we should keep testing it.  The web worker test is trickier to deal with so I just omitted it while doing browserify tests.  Can call that good enough for now.

davidlehn:
Merged via https://github.com/digitalbazaar/forge/pull/645. Thanks!

leoselig:
@davidlehn Fair enough. Just to clarify though, I was not suggesting to break node-forge for browserify users. I only wanted to remove it from the test execution since it seemed effectively irrelevant what bundler is used for the tests.

Thanks for merging this and for all the work on this package! :)

Issue #279, opened 2015-07-07
Detached PKCS#7 signature


BaurzhanSakhariev:
Will be there any parameter to generate attached or detached PKCS#7 signature? It would be very useful. 

I mean something like this:

```
signature = createPkcs7(data, false) // get pure signature, without data
signature = createPkcs7(data,true)  // get pkcs7 signature with attached data
```

dlongley:
I haven't looked into it yet, but, naively, it does seem like it would be fairly easy to support detached signatures. We'll keep this issue open as a desirable future feature; PRs welcome.

fadlijuniarso:
@BaurzhanSakhariev you can check this one: https://gitlab.com/rootcaid/kominfo-pki-websdk/blob/master/lib/forge/pkcs7.js#L331

vbuch:
@dlongley The changes propsed by @fadlijuniarso 3 years ago: https://github.com/digitalbazaar/forge/pull/605

davidlehn:
Thanks to @vbuch and others, the feature was added in 0.7.6:
https://github.com/digitalbazaar/forge/blob/master/CHANGELOG.md
https://github.com/digitalbazaar/forge#pkcs7

PR #605, opened 2018-08-08
Add sign({detached: true}) to PKCS#7

vbuch:
Adds `signDetached()` to the PKCS#7 implementation (see #279) as in my usecase it worked perfectly fine the way it was proposed 3 years ago by @fadlijuniarso. The same piece of code is being used in a number of repos on GitHub but what lead me here was the piece used in [pdfsign.js](https://github.com/Communication-Systems-Group/pdfsign.js/blob/master/src/lib/pkcs7-detached.js#L324) so thanks to @tbocek as well.
I've also added a test to cover `signDetached()`.

### DIFF
--- a/lib/pkcs7.js
+++ b/lib/pkcs7.js
@@ -332,7 +332,7 @@ p7.createSignedData = function() {
      * Includes the signed data with digital signature.
      */
     sign: function() {
-      msg._baseSign(false);
+      msg._baseSign();
     },
 
     /**
@@ -341,14 +341,16 @@ p7.createSignedData = function() {
      * Includes the signature and certificate without the signed data.
      */
     signDetached: function() {
-      msg._baseSign(true);
+      msg._baseSign({detached: true});
     },
 
     /**
      * Signs the content.
-     * @param detached If signing should be done in detached mode
+     * @param options Options to apply when signing:
+     *    [detached] boolean. If signing should be done in detached mode. Defaults to false.
      */
-    _baseSign: function(detached) {
+    _baseSign: function(options) {
+      options = options || {};
       // auto-generate content info
       if(typeof msg.content !== 'object' || msg.contentInfo === null) {
         // use Data ContentInfo
@@ -368,7 +370,7 @@ p7.createSignedData = function() {
             content = forge.util.encodeUtf8(msg.content);
           }
 
-          if (detached) {
+          if (options.detached) {
             msg.detachedContent = asn1.create(asn1.Class.UNIVERSAL, asn1.Type.OCTETSTRING, false, content);
           } else {
             msg.contentInfo.value.push(
--- a/lib/pkcs7.js
+++ b/lib/pkcs7.js
@@ -332,7 +332,7 @@ p7.createSignedData = function() {
      * Includes the signed data with digital signature.
      */
     sign: function() {
-      msg.baseSign(false);
+      msg._baseSign(false);
     },
 
     /**
@@ -341,13 +341,14 @@ p7.createSignedData = function() {
      * Includes the signature and certificate without the signed data.
      */
     signDetached: function() {
-      msg.baseSign(true);
+      msg._baseSign(true);
     },
 
     /**
      * Signs the content.
+     * @param detached If signing should be done in detached mode
      */
-    baseSign: function(detached) {
+    _baseSign: function(detached) {
       // auto-generate content info
       if(typeof msg.content !== 'object' || msg.contentInfo === null) {
         // use Data ContentInfo
@@ -462,17 +463,18 @@ p7.createSignedData = function() {
     var content;
 
     if (msg.detachedContent) {
+      // Signature has been made in detached mode.
       content = msg.detachedContent;
     } else {
+      // Note: ContentInfo is a SEQUENCE with 2 values, second value is
+      // the content field and is optional for a ContentInfo but required here
+      // since signers are present
       // get ContentInfo content
       content = msg.contentInfo.value[1];
       // skip [0] EXPLICIT content wrapper
       content = content.value[0];
     }
 
-    // Note: ContentInfo is a SEQUENCE with 2 values, second value is
-    // the content field and is optional for a ContentInfo but required here
-    // since signers are present
     if(!content) {
       throw new Error(
         'Could not sign PKCS#7 message; there is no content to sign.');
--- a/lib/pkcs7.js
+++ b/lib/pkcs7.js
@@ -328,8 +328,26 @@ p7.createSignedData = function() {
 
     /**
      * Signs the content.
+     *
+     * Includes the signed data with digital signature.
      */
     sign: function() {
+      msg.baseSign(false);
+    },
+
+    /**
+     * Signs the content in detached mode
+     *
+     * Includes the signature and certificate without the signed data.
+     */
+    signDetached: function() {
+      msg.baseSign(true);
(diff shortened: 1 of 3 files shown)


tbocek:
awesome!

lib/pkcs7.js
```diff
@@ -328,8 +328,26 @@ p7.createSignedData = function() {
...
+     * Signs the content in detached mode
+     *
+     * Includes the signature and certificate without the signed data.
+     */
+    signDetached: function() {
+      msg.baseSign(true);
+    },
+
+    /**
+     * Signs the content.
```
mattcollier: please add `@param` etc to the function docs.

lib/pkcs7.js
```diff
@@ -328,8 +328,26 @@ p7.createSignedData = function() {
...
+     */
+    baseSign: function(detached) {
```
mattcollier: Should this be `_baseSign`?  Do we want this API to be part of the public API?

lib/pkcs7.js
```diff
@@ -437,22 +459,28 @@ p7.createSignedData = function() {
...
+      content = msg.contentInfo.value[1];
+      // skip [0] EXPLICIT content wrapper
+      content = content.value[0];
+    }
+
     // Note: ContentInfo is a SEQUENCE with 2 values, second value is
     // the content field and is optional for a ContentInfo but required here
     // since signers are present
-    if(msg.contentInfo.value.length < 2) {
+    if(!content) {
```
mattcollier: since `contentInfo` is no longer being used here, the above comment becomes irrelevant.  Can it be moved to some other useful location?

lib/pkcs7.js
vbuch: Fixed in https://github.com/digitalbazaar/forge/pull/605/commits/09afde45c47fa21f58038bf1685ff06b9410ceb2

lib/pkcs7.js
vbuch: Fixed in https://github.com/digitalbazaar/forge/pull/605/commits/09afde45c47fa21f58038bf1685ff06b9410ceb2

lib/pkcs7.js
vbuch: Fixed in https://github.com/digitalbazaar/forge/pull/605/commits/09afde45c47fa21f58038bf1685ff06b9410ceb2

lib/pkcs7.js
```diff
@@ -341,13 +341,14 @@ p7.createSignedData = function() {
...
-      msg.baseSign(true);
+      msg._baseSign(true);
     },
 
     /**
      * Signs the content.
+     * @param detached If signing should be done in detached mode
      */
-    baseSign: function(detached) {
+    _baseSign: function(detached) {
```
mattcollier: @vbuch I don't know what the future holds for `_baseSign`.  I really don't like the look of `msg._baseSign(true)`.  How do you feel about the signature being `msg._baseSign({detached: true})`, and for the default case, it can just be `msg._baseSign()`?

vbuch: Would introduce a couple of lines of more code, but would look much better, yes. Will do it.

vbuch:
@davidlehn Agreed. I'm just used to gitmoji.

lib/pkcs7.js
vbuch: Hm... I'm thinking now that I see the code... baseSign could be needless in this case. As long as sing() has an optional `options` parameter we could call sign() and sign({detached: true}) instead of sign() and signDetached() What do you think @mattcollier ?

mattcollier: @vbuch good point.  It will be up to @dlongley and @davidlehn to answer this question.

davidlehn:
@vbuch Huh, I didn't realize gitmoji was a "spec".  I just thought emoji were random and people doing that were crazy. :-)  It's not a bad idea I guess, but kind of weird to just use it for a few commits.  Also seems a "spec" mapping keywords like ":bugfix:" to emoji would make lots more sense.  But I'm getting off topic.  Could just rewrite the commit message history on your branch and force push.  Or we could just not worry about it.  Sorry for the hassle and thanks for the patch.

vbuch:
@davidlehn Will do that. And what about this https://github.com/digitalbazaar/forge/pull/605#discussion_r209168056

davidlehn:
@vbuch: I don't have strong opinions on the api, mainly because I don't know what detached signatures are used for! ;-)  @dlongley?  I'd say just have `sign(options)`, and if people get confused we can add the helper function `signDetached = () => sign({detached: true})`.

davidlehn:
@vbuch Was the test case generated by some other app or by the new code itself?  Just wondering where to look to see if the detached sig construction is correct.  Not quite sure where spec text is for that.

vbuch:
What I am using detached signatures for is signing PDFs. The one in the test is generated by the code. I'm working on an actual test for its validity here https://github.com/vbuch/node-signpdf Once I'm done I could possibly have better testing for the code here thou I'm not sure more is really needed.

davidlehn:
Ok.  Testing against itself is good for regressions, but hard to say if it's compatible with other tools.  I'm not keen on reading that whole adobe pdf doc about this stuff. ;-)  Once that options api issue is addressed will likely just merge and release this.  Can fix and/or improve testing later if there are issues.

vbuch:
@davidlehn @mattcollier @dlongley I think I did all the things we agreed on. CI passed. Seems to me this piece is good to go. 🎉

dirkx:
@vbuch A bit of an odd issue - using your detached:true I am still seeing the payload in the PKCS#7.

I.e running ` node x.js | openssl pkcs7 -print -text` shows the `contents: type: pkcs7-data`  field in the output.

```
var forge = require('node-forge');
var fs = require('fs');


// openssl req -x509 -subj /CN=foo -keyout x.key -out x.crt -nodes
var certAsPem = fs.readFileSync('x.crt');
var cert = forge.pki.certificateFromPem(certAsPem)

var keyAsPem = fs.readFileSync('x.key');
var privateKey = forge.pki.privateKeyFromPem(keyAsPem);

var p7 = forge.pkcs7.createSignedData();

p7.content = forge.util.createBuffer('TEK:12321378, TEK:1278391287189, TEK:12321378, TEK:1278391287189,TEK:12321378, TEK:1278391287189,TEK:12321378, TEK:1278391287189,TEK:12321378, TEK:1278391287189,TEK:12321378, TEK:1278391287189', 'utf8');
p7.addCertificate(cert);

p7.addSigner({
  key: privateKey,
  certificate: cert,
  digestAlgorithm: forge.pki.oids.sha256,
  authenticatedAttributes: [{
    type: forge.pki.oids.contentType,
    value: forge.pki.oids.data
  }, {
    type: forge.pki.oids.messageDigest
    // value will be auto-populated at signing time
  }, {
    type: forge.pki.oids.signingTime,
    // value can also be auto-populated at signing time
    value: new Date()
  }]
});
p7.sign();

var pem = forge.pkcs7.messageToPem(p7);

// PKCS#7 Sign in detached mode.
// Includes the signature and certificate without the signed data.
//
p7.sign({detached: true});

var out = forge.pkcs7.messageToPem(p7);

console.log(out);
```

vbuch:
Sorry @dirkx. I won't be able to help. I see the lib has gone 2 versions ahead since my addition was released. Maybe something changed meanwhile n

PR #614, opened 2018-09-26
Add description (2.5.4.13) to OID list.

herpiko:


davidlehn:
Is `description` common in certs?

There are lots of [2.5.4 OIDs](http://oidref.com/2.5.4) and I'm unsure how the ones that are predefined were choosen.  I think it's based on common and expected usage.  I'm fine with adding common OIDs but we of course can't add everything.  @dlongley Thoughts on this?

herpiko:
@davidlehn 

I thought it isn't common until I found that Indonesia govt is using this attribute in their PKI implementation, including CSR generation in web client side. It may contains additional information about the certificate owner. They're also using https://gitlab.com/rootcaid/kominfo-pki-websdk that based on forge.

```
➜  Downloads openssl x509 -in dev\ vel.pem -text | grep description
        Subject: emailAddress=censoredcensored@gmail.com, CN=dev vel, OU=Tim Teknis, O=Balai Sertifikasi Elektronik, L=Jakarta, ST=DKI Jakarta, C=ID/description=123456789_Tanda Tangan Digital Demo
```

PR #616, opened 2018-10-09
Forge: Removed phantomjs dependency

ossdev07:
PhantomJS is an unmaintained project, removed phantomjs support from package.json.
Changed browser to chromeheadless in karma.conf.js, Updated jshint version to remove optional phantomjs dependency.

Signed-off-by: ossdev07 <ossdev@puresoftware.com>


ossdev07:
Hi,

Added custom launcher support for headless chrome.
-> Reason for using --no-sandbox flag:

For aarch64 chrome binary is not present, we are using chromium binary to launch the Chrome.
Security is one of the most important goals for Chromium, and --no-sandbox flag provides that.
When trying to use chromium as root user like in docker container then only the requirement of --no-sandbox flag is coming
Without the flag, chromium gives error as root user-
~~~
[28769:28769:0925/163702.505790:ERROR:zygote_host_impl_linux.cc(89)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180.
~~~
Thus, we need to give the --no-sandbox flag when we launch chrome with chromium binary.

ossdev07:
Ping @dlongley @davidlehn 

Please have a look at this PR.
It will be a delight to see phantomjs removed from the package.

dlongley:
@davidlehn, I'm +1 for removing PhantomJS. If you approve, please merge.

package.json
```diff
@@ -20,7 +20,7 @@
     "cross-env": "^5.1.3",
     "express": "^4.16.2",
     "jscs": "^3.0.7",
-    "jshint": "^2.9.5",
+    "jshint": "^0.8.0",
```
dlongley: Is this necessary?

dlongley: I see that you said:

> Updated jshint version to remove optional phantomjs dependency.

Seems odd that we'd have to go back 2 major versions just to remove PhantomJS.

davidlehn: jscs should be dropped in favor of eslint.  probably should just drop jshint too in favor of eslint.

davidlehn:
I'm not sure I understand the `--no-sandbox` part.  Do we want to always use that?

Part of the reason for using phantomjs was it was so out-of-date with standards.  Not sure how we can test code works in old browsers.  Running headless Chrome or Firefox isn't much different than testing in Node.js from an API perspective.

davidlehn:
Also get permission denied trying to view that chromium bug

package.json
ossdev07: As I mentioned earlier I did this change only to remove opional phantomjs dependency, but I agree with you that instead of dropping jshint version we should drop jshint and jscs from the  package.json.
I will be happy to update this PR accordingly if you agree.

ossdev07:
Hi @davidlehn 

> I'm not sure I understand the `--no-sandbox` part. Do we want to always use that?

-> Reason for using --no-sandbox flag:

For aarch64 chrome binary is not present, we are using chromium binary to launch the Chrome.
Security is one of the most important goals for Chromium, and --no-sandbox flag provides that.
When trying to use chromium as root user like in docker container then only the requirement of --no-sandbox flag is coming
Without the flag, chromium gives error as root user-
```
[28769:28769:0925/163702.505790:ERROR:zygote_host_impl_linux.cc(89)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180.
```

Thus, we need to give the --no-sandbox flag when we launch chrome with chromium binary.

Please do revert if you still find anything not clear.

davidlehn:
PR to just remove PhantomJS: https://github.com/digitalbazaar/forge/pull/644

ossdev07:
@davidlehn Thanks for considering this PR, I have looked #644 I think it will be good to remove the phantomjs dependency for now.

davidlehn:
PhantomJS removed in https://github.com/digitalbazaar/forge/pull/644.  Please update or close this PR as needed.

ossdev07:
Thanks for accepting the changes suggested by this PR @davidlehn.
I am closing this pull request now.

PR #620, opened 2018-10-17
 Add encodeBase64Url/decodeBase64Url from did-io

dmitrizagidulin:
Move the `encodeBase64Url`/`decodeBase64Url` from `did-io` (since they're using forge anyway).


lib/util.js
```diff
@@ -1627,6 +1644,32 @@ util.decode64 = function(input) {
...
+ */
+util.decodeBase64Url = function decodeBase64Url(input) {
+  let normalInput = input
+    .replace(/-/g, '+')
+    .replace(/_/g, '/');
+  const mod4 = normalInput.length % 4;
+  if(mod4 === 0) {
+    // pass
+  } else if(mod4 === 2) {
+    normalInput = normalInput + '==';
```
mattcollier:
```suggestion
    normalInput += '==';
```

lib/util.js
```diff
@@ -1627,6 +1644,32 @@ util.decode64 = function(input) {
...
+  } else if(mod4 === 3) {
+    normalInput = normalInput + '=';
```
mattcollier:
```suggestion
    normalInput += '=';
```

lib/util.js
```diff
@@ -1627,6 +1644,32 @@ util.decode64 = function(input) {
...
+ * @param input the data to decode.
+ *
+ * @return the decoded value.
+ */
+util.decodeBase64Url = function decodeBase64Url(input) {
+  let normalInput = input
+    .replace(/-/g, '+')
+    .replace(/_/g, '/');
+  const mod4 = normalInput.length % 4;
+  if(mod4 === 0) {
```
mattcollier: it appears that `mod4 === 1` is the error condition, maybe we can do without the === 0 check and be explicit about the error condition.

package.json
```diff
@@ -115,6 +115,12 @@
   "jspm": {
     "format": "amd"
   },
+  "eslintConfig": {
```
mattcollier: we've been using `.eslintrc` files which I think give a bit more granularity.  So there could be one at the root with 
```
{env: {node: true}} // does this need browser as well?
```
and then in the test folder, you can do
```
{env: {mocha: true}) // only, since .eslintrc file are additive
```

lib/util.js
dmitrizagidulin: Sure!

lib/util.js
dmitrizagidulin: Fixing.

lib/util.js
dmitrizagidulin: Fixing.

package.json
dmitrizagidulin: Sure. I'm not quite familiar with those files, let me try the suggestion above.

dmitrizagidulin: Added

lib/util.js
```diff
@@ -1590,6 +1590,23 @@ util.encode64 = function(input, maxline) {
   return output;
 };
 
+/**
+ * Encodes input according to the "Base64url Encoding" format as specified
+ * in JSON Web Signature (JWS) RFC7517. A URL safe character set is used and
+ * trailing '=', line breaks, whitespace, and other characters are omitted.
+ *
+ * @param input the data to encode.
```
dlongley: Can we specify what format `data` is in? Is it a binary string? I would assume so.

lib/util.js
```diff
@@ -1627,6 +1644,30 @@ util.decode64 = function(input) {
...
 };
 
+/**
+ * Decodes input according to the "Base64url Encoding" format as specified
+ * in JSON Web Signature (JWS) RFC7517. A URL safe character set is used and
+ * trailing '=', line breaks, whitespace, and other characters are omitted.
+ *
+ * @param input the data to decode.
+ *
+ * @return the decoded value.
```
dlongley: Can we say that the output is a binary string?

dmitrizagidulin: Sure. Fixing.

lib/util.js
dmitrizagidulin: Yep, fixing.

.eslintrc
```diff
@@ -0,0 +1 @@
+{env: {node: true}}
```
mattcollier: eslintrc files are JSON: https://eslint.org/docs/user-guide/configuring

dmitrizagidulin:
> I suppose this is ok for the binary string version

I'm merely moving the code from did-io :)

.eslintrc
dmitrizagidulin: Whoops, fixing.

tests/unit/util.js
```diff
@@ -234,6 +234,18 @@ var UTIL = require('../../lib/util');
       ASSERT.equal(UTIL.decode64(s2), s1);
     });
 
+    it('should base64Url encode some bytes', function() {
```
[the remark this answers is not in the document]
dmitrizagidulin: Should be doable? We can also just scrap this PR and use an external library, like https://www.npmjs.com/package/base64url :)

mattcollier:
@dlongley any objections to just using https://www.npmjs.com/package/base64url in did-io?

dlongley:
@mattcollier,

> any objections to just using https://www.npmjs.com/package/base64url in did-io?

Won't solve the problem when we make did-io function in the browser (which we need to do).

dmitrizagidulin:
@dlongley 
> Won't solve the problem when we make did-io function in the browser (which we need to do).

Fwiw, I've used the `base64url` lib in the browser before. It's not large, has no dependencies, and Webpacks just fine.

dlongley:
@dmitrizagidulin,

Looks like it uses a Buffer polyfill. That's what we'd be trying to avoid. If it used Uint8Array then it would actually have no dependencies. I think I wrote some code that does that somewhere? We can turn that into a module and use that instead if that's the direction we want to go.

dlongley:
https://github.com/digitalbazaar/bedrock-web-private-remote-storage/blob/implementation/base64url.js

dmitrizagidulin:
Ok, per discussion on IRC, closing this PR for now. I can use the `base64url` package in did-io meanwhile, and will open an issue to replace it with something non-Buffer-polyfill-dependent.

Issue #530, opened 2017-08-09
x509 certificate as PKCS #12 never valid if notAfter is in the year 2050 or later


bes:
First of all, thanks for a great project, with good documentation & examples.

I am trying to create a PKCS12 archive from a keypair and a self-signed certificate like so:

```js
forge.rsa.generateKeyPair({bits: 1024, workers: -1}, (err, keypair) => {
        if (err) {
            console.log(err);
            return;
        }
        const privateKey = keypair.privateKey;
        const publicKey = keypair.publicKey;

        const pki = forge.pki;

        const cert = pki.createCertificate();
        cert.publicKey = publicKey;
        cert.serialNumber = '01';
        const attrs = [
            {
                name: 'commonName',
                value: 'x',
            },
            {
                name: 'countryName',
                value: 'x',
            },
            {
                shortName: 'ST',
                value: 'x',
            },
            {
                name: 'localityName',
                value: 'x',
            },
            {
                name: 'organizationName',
                value: 'x',
            },
            {
                shortName: 'OU',
                value: 'x',
            },
        ];
        cert.setSubject(attrs);
        cert.setIssuer(attrs);

        cert.validity.notBefore = new Date();
        cert.validity.notAfter = new Date();
        // Valid for 100 years
        cert.validity.notAfter.setFullYear(cert.validity.notBefore.getFullYear() + 100);
        cert.sign(privateKey);

        const p12 = forge.pkcs12.toPkcs12Asn1(privateKey, cert, 'somePassword', {algorithm: '3des'});
        const p12Der = forge.asn1.toDer(p12).getBytes();
        let p12b64 = forge.util.encode64(p12Der);

        console.log(p12b64);

        // Copy the Base64 and decode as binary into a .p12-file
    });
});
```

That certificate is not valid since the date is after 2050, and not encoded as a GeneralizedTime as stated in [RFC2459](https://www.ietf.org/rfc/rfc2459.txt)

> certificate validity dates in 2050 or later MUST be encoded as GeneralizedTime.

Is it possible to create a certificate with GeneralizedTime in some way? Why isn't the date automatically encoded as GeneralizedTime if the date is on or after the year 2050?

Thanks!

davidlehn:
[...]

Issue #12, opened 2013-01-01
Using certificate signing requests


mrboggieman:
Is there a way I could generate a certificate signing request (based on PKCS#10) using a locally generated key?

PR #634, opened 2019-01-02
Create certificates with not before/after years greater or equal 2050

sruetzler:
Should fix issue:
x509 certificate as PKCS #12 never valid if notAfter is in the year 2050 or later #530

### DIFF
--- a/lib/x509.js
+++ b/lib/x509.js
@@ -2474,6 +2474,8 @@ function _CRIAttributesToAsn1(csr) {
  */
 pki.getTBSCertificate = function(cert) {
   // TBSCertificate
+  var notBefore = dateToPem(cert.validity.notBefore);
+  var notAfter = dateToPem(cert.validity.notAfter);
   var tbs = asn1.create(asn1.Class.UNIVERSAL, asn1.Type.SEQUENCE, true, [
     // version
     asn1.create(asn1.Class.CONTEXT_SPECIFIC, 0, true, [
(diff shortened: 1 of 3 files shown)


tests/unit/x509.js
```diff
@@ -406,6 +471,49 @@ var UTIL = require('../../lib/util');
...
+        name: 'localityName',
+        value: 'Blacksburg'
+      }, {
+        name: 'organizationName',
+        value: 'Test'
+      }, {
+        shortName: 'OU',
+        value: 'Test'
+      }];
+      let notBefore = new Date("2050-02-02");
```
davidlehn: Use 'var' to fix the test suite.  We'll eventually move everything to ES6+ in one huge change.

tests/unit/x509.js
```diff
@@ -406,6 +471,49 @@ var UTIL = require('../../lib/util');
...
+        subject: attrs,
+        issuer: attrs,
+        isCA: true,
+        notBefore : notBefore
+      });
+
+      var pem = PKI.certificateToPem(cert);
+      cert = PKI.certificateFromPem(pem);
+
+      let notAfter = new Date("2051-02-02");
```
davidlehn: Use 'var'.

lib/x509.js
```diff
@@ -2543,6 +2541,21 @@ pki.getTBSCertificate = function(cert) {
   return tbs;
 };
 
+const jan_1_1950 = new Date('1950-01-01T00:00:00Z');
+const jan_1_2050 = new Date('2050-01-01T00:00:00Z');
+
+function dateToPem(date){
```
davidlehn: This is going to ASN1.  Should rename it to `_dateToAsn1` and perhaps move it near one of the other `_*ToAsn1` helper functions.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "node-forge",
-  "version": "0.7.7-dev",
+  "version": "0.7.6-1",
```
davidlehn: Would rather not have this in the patch since we try to keep the -dev suffix for unreleased code.

sruetzler:
I close the pull request and open another with the requested changes

PR #635, opened 2019-01-07
Custom validity period verification date

ahoppen:
## Problem
I would like to be able to check whether a particular certificate chain was valid at a specific date at which the certificate was used to sign a message. 

## Solution
For this I added an additional parameter to `x509.verifyCertificateChain` with which the date at which the certificate's validity shall be checked can be specified. It defaults to the current date.

## Open questions
Passing `validityCheckDate` after the `verify` callback is a bit unfortunate, but adding it before the callback would be a breaking change and I'm not sure how strict you are with breaking changes at this point.

### DIFF
--- a/lib/x509.js
+++ b/lib/x509.js
@@ -2940,7 +2940,13 @@ pki.certificateError = {
  * @param caStore a certificate store to verify against.
  * @param chain the certificate chain to verify, with the root or highest
  *          authority at the end (an array of certificates).
- * @param verify called for every certificate in the chain.
+ * @param options a callback to be called for every certificate in the chain or 
+ *                  an object with:
+ *                  verify a callback to be called for every certificate in the
+ *                    chain
+ *                  validityCheckDate the date against which the certificate 
+ *                    validity period should be checked. Pass null to not check 
+ *                    the validity period. By default, the current date is used. 
  *
  * The verify callback has the following signature:
  *
@@ -2956,7 +2962,7 @@ pki.certificateError = {
  *
  * @return true if successful, error thrown if not.
  */
-pki.verifyCertificateChain = function(caStore, chain, verify) {
+pki.verifyCertificateChain = function(caStore, chain, options) {
   /* From: RFC3280 - Internet X.509 Public Key Infrastructure Certificate
     Section 6: Certification Path Validation
     See inline parentheticals related to this particular implementation.
@@ -3086,14 +3092,27 @@ pki.verifyCertificateChain = function(caStore, chain, verify) {
        CAs that may appear below a CA before only end-entity certificates
        may be issued. */
 
+  // if a verify callback is passed as the third parameter, package it within 
+  // the options object. This is to support a legacy function signature that 
+  // expected the verify callback as the third parameter.
+  if(typeof options === 'function') {
+    options = {verify: options};
+  }
+  options = options || {};
+
   // copy cert chain references to another array to protect against changes
   // in verify callback
   chain = chain.slice(0);
   var certs = chain.slice(0);
 
-  // get current date
-  var now = new Date();
-
+  var validityCheckDate = options.validityCheckDate;
+  // if no validityCheckDate is specified, default to the current date. Make 
+  // sure to maintain the value null because it indicates that the validity 
+  // period should not be checked.
+  if(typeof validityCheckDate === 'undefined') {
+    validityCheckDate = new Date();  
+  }
+  
   // verify each cert in the chain using its parent, where the parent
   // is either the next in the chain or from the CA store
   var first = true;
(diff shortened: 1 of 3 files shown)


davidlehn:
Yes, it's difficult to add optional arguments after optional arguments.  A better style when that is an issue is to have an options object for all such arguments, or even all arguments.  I imagine the forge APIs would all move that direction if we ever had time to do major API updates.  I don't think we should break the API now if possible.  We could hack on an options object approach but maybe this is rare enough it's not worth the trouble?

davidlehn:
Maybe I misread the API about which args are optional.  In any case, an options object would likely be nicer.

ahoppen:
Thanks for the feedback. I didn't realise the issue had already come up that often.

## Changes since the last revision
1. Changed the third parameter to be an options object that takes the `validityCheckDate` and `verifyCallback`. If a function is passed per the old function signature, it is used as the `verifyCallback`. We thus maintain API compatibility.
2. It appears that not checking the validity period is also a requested feature. While I was at it, I changed the behaviour so the validity period is not checked if `null` is passed as the `validityCheckDate`

lib/x509.js
```diff
@@ -3074,15 +3093,18 @@ pki.verifyCertificateChain = function(caStore, chain, verify) {
...
+    if(validityCheckDate) {
+      // 1. check valid time
+      if(validityCheckDate < cert.validity.notBefore || 
+         validityCheckDate > cert.validity.notAfter) {
+        error = {
+          message: 'Certificate is not valid yet or has expired.',
+          error: pki.certificateError.certificate_expired,
+          notBefore: cert.validity.notBefore,
+          notAfter: cert.validity.notAfter,
+          now: validityCheckDate
```
ahoppen: We should actually rename this from `now` to `validityCheckDate`. However, again, I'm not sure how strict you are about breaking the API. The options I see here are to
1. Rename `now` to `validityCheckDate`
2. Add a second field `validityCheckDate` and keep `now` for legacy reasons
3. Stick with `now` even though it is incorrect

What do you think?

davidlehn: We would rather not change the API.  I think `now` is ok.  It's what you specified as "now".  Could leave a comment to revisit the name in the future if larger changes are happening.

lib/x509.js
```diff
@@ -2910,7 +2910,13 @@ pki.certificateError = {
...
  * @param chain the certificate chain to verify, with the root or highest
  *          authority at the end (an array of certificates).
- * @param verify called for every certificate in the chain.
+ * @param options a callback to be called for every certificate in the chain or 
+ *                  an object with:
+ *                  verifyCallback a callback to be called for every certificate 
+ *                    in the chain
+ *                  validityCheckDate the date against which the certificate 
+ *                    validity period should be checked. Pass null to not check 
+ *                    the validity period.By default, the current date is used.
```
davidlehn: Add a space:
```suggestion
 *                    the validity period. By default, the current date is used. 
```

lib/x509.js
```diff
@@ -2926,7 +2932,7 @@ pki.certificateError = {
  *
  * @return true if successful, error thrown if not.
  */
-pki.verifyCertificateChain = function(caStore, chain, verify) {
+pki.verifyCertificateChain = function(caStore, chain, options) {
```
davidlehn: This is the direction we'd go in the future, but it will change the API.  The signature should be `(caStore, chain, verify, options)` for now.  That leaves API open to other options as needed without breaking anything.

davidlehn:
I merged in https://github.com/digitalbazaar/forge/pull/636, which caused a conflict here.  Sorry about that.  Hopefully not too hard to fix up.

ahoppen:
No worries about merging the other PR. The conflict was easy to fix. I incorporated the requested changes. 
IMHO I did not perform an API change though because we still fall back to the old behaviour if a function is passed as the third parameter.

lib/x509.js
davidlehn: Ah, I missed that part.  That's fine then.

lib/x509.js
```diff
@@ -2931,7 +2931,13 @@ pki.certificateError = {
  * @param caStore a certificate store to verify against.
  * @param chain the certificate chain to verify, with the root or highest
  *          authority at the end (an array of certificates).
- * @param verify called for every certificate in the chain.
+ * @param options a callback to be called for every certificate in the chain or 
+ *                  an object with:
+ *                  verifyCallback a callback to be called for every certificate
```
davidlehn: Can you rename this to just "verify"?  I don't think we need the "Callback" suffix.

ahoppen: Done

ahoppen:
Just a friendly ping. Is there anything else you'd like me to change on this PR?

davidlehn:
I just haven't had time to get to forge PRs.  It's next up.  Perhaps you'd be interested in looking at also pulling in the TLS bit from #386?  I was waiting to merge this PR until I had looked at how to fix up that patch too.  That one used a trailing options param which made it easier to just pass through options from the TLS code. Need to figure out how to do something similar with the code in this PR.

ahoppen:
No worries, I just know how these things can get lost sometimes.

I have incorporated the TLS changes from #386 into this PR. Could you take a look?

PR #639, opened 2019-01-13
Use node.crypto.generateKeyPair when available

kevincennis:
Node >= 10.12.0 has a native `crypto.generateKeyPair` that's quite a bit faster (like, 10-20x) than pure JS.

Apologies if this PR is missing something. This was my first time digging into Forge.

### DIFF
--- a/lib/rsa.js
+++ b/lib/rsa.js
@@ -894,7 +894,7 @@ pki.rsa.generateKeyPair = function(bits, e, options, callback) {
         modulusLength: bits,
         publicExponent: e,
         publicKeyEncoding: {
-          type: 'pkcs1',
+          type: 'spki',
           format: 'pem'
         },
         privateKeyEncoding: {
--- a/lib/rsa.js
+++ b/lib/rsa.js
@@ -905,10 +905,9 @@ pki.rsa.generateKeyPair = function(bits, e, options, callback) {
         if (err) {
           return callback(err);
         }
-        var privateKey = pki.privateKeyFromPem(priv);
         callback(null, {
-          privateKey: privateKey,
-          publicKey: pki.setRsaPublicKey(privateKey.n, privateKey.e)
+          privateKey: pki.privateKeyFromPem(priv),
+          publicKey: pki.publicKeyFromPem(pub)
         });
       });
     }
--- a/lib/rsa.js
+++ b/lib/rsa.js
@@ -74,6 +74,8 @@ if(typeof BigInteger === 'undefined') {
   var BigInteger = forge.jsbn.BigInteger;
 }
 
+var _crypto = forge.util.isNodejs ? require('crypto') : null;
+
 // shortcut for asn.1 API
 var asn1 = forge.asn1;
 
@@ -887,6 +889,30 @@ pki.rsa.generateKeyPair = function(bits, e, options, callback) {
   // key generation code if available and if parameters are acceptable
   if(!forge.options.usePureJavaScript && callback &&
     bits >= 256 && bits <= 16384 && (e === 0x10001 || e === 3)) {
+    if(forge.util.isNodejs && _detectNodeCrypto('generateKeyPair')) {
+      return _crypto.generateKeyPair('rsa', {
+        modulusLength: bits,
+        publicExponent: e,
+        publicKeyEncoding: {
+          type: 'pkcs1',
+          format: 'der'
+        },
+        privateKeyEncoding: {
+          type: 'pkcs8',
+          format: 'der'
+        }
+      }, function(err, pub, pkcs8) {
+        if (err) {
+          return callback(err);
+        }
+        var privateKey = pki.privateKeyFromAsn1(
+          asn1.fromDer(forge.util.createBuffer(pkcs8)));
+        callback(null, {
+          privateKey: privateKey,
+          publicKey: pki.setRsaPublicKey(privateKey.n, privateKey.e)
+        });
+      });
+    }
     if(_detectSubtleCrypto('generateKey') && _detectSubtleCrypto('exportKey')) {
       // use standard native generateKey
       return window.crypto.subtle.generateKey({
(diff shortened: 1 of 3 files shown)


lib/rsa.js
```diff
@@ -887,6 +889,30 @@ pki.rsa.generateKeyPair = function(bits, e, options, callback) {
   // key generation code if available and if parameters are acceptable
   if(!forge.options.usePureJavaScript && callback &&
     bits >= 256 && bits <= 16384 && (e === 0x10001 || e === 3)) {
+    if(forge.util.isNodejs && _detectNodeCrypto('generateKeyPair')) {
```
mattcollier: I wonder if this conditional might be simplified.  if the `util.isNodejs` call is necessary, maybe it could go into `detectNodeCrypto`.

lib/rsa.js
```diff
@@ -887,6 +889,30 @@ pki.rsa.generateKeyPair = function(bits, e, options, callback) {
...
+        },
+        privateKeyEncoding: {
+          type: 'pkcs8',
+          format: 'der'
+        }
+      }, function(err, pub, pkcs8) {
+        if (err) {
+          return callback(err);
+        }
+        var privateKey = pki.privateKeyFromAsn1(
```
mattcollier: Is there a shorter route from whatever format `generateKeyPair` might generate to a `privateKey`?

lib/rsa.js
kevincennis: Yep. Will do.

lib/rsa.js
kevincennis: Looks like I can ask Node for a `pem` and then use `pki.privateKeyFromPem()`.

lib/rsa.js
```diff
@@ -887,6 +889,29 @@ pki.rsa.generateKeyPair = function(bits, e, options, callback) {
...
+          format: 'pem'
+        }
+      }, function(err, pub, priv) {
+        if (err) {
+          return callback(err);
+        }
+        var privateKey = pki.privateKeyFromPem(priv);
+        callback(null, {
+          privateKey: privateKey,
+          publicKey: pki.setRsaPublicKey(privateKey.n, privateKey.e)
```
mattcollier: Thanks again, can `pub` be used directly somehow here?  I'm comparing this implementation to another recent bit of dev code.  I notice that `publicKeyEncoding.type` is `spki`.  Is that relevant here?

https://github.com/digitalbazaar/crypto-ld/blob/signer/lib/ld-key-pair.js#L357-L376

lib/rsa.js
```diff
@@ -1727,6 +1752,17 @@ function _getMillerRabinTests(bits) {
...
 }
 
+/**
+ * Performs feature detection on the Node crypto interface.
+ *
+ * @param fn the feature (function) to detect.
+ *
+ * @return true if detected, false if not.
+ */
+function _detectNodeCrypto(fn) {
```
mattcollier: I suggest calling this `_detectNodeCryptoMethod`.

lib/rsa.js
mattcollier: Does `pki.publicKeyFromPem` do the right thing? https://github.com/digitalbazaar/forge/blob/26bf82aed3b9814ab032cb88244b8176d6653bb1/lib/x509.js#L751-L752

kevincennis: Tests pass.

Guess I should've grepped the whole project for `publicKeyFromPem`. Just assumed it didn't exist because it wasn't next to `privateKeyFromPem` in `pki.js`.

Will update.

lib/util.js
```diff
@@ -125,6 +125,11 @@ util.isArrayBufferView = function(x) {
   return x && util.isArrayBuffer(x.buffer) && x.byteLength !== undefined;
 };
 
+// Node versions >= 10.2 support native keyPair generation,
```
davidlehn: Docs say it's 10.12.0:
https://nodejs.org/dist/latest-v10.x/docs/api/crypto.html#crypto_crypto_generatekeypair_type_options_callback

kevincennis: Woof. Good catch. Will update.

kevincennis: 👋

lib/rsa.js
davidlehn: Was this going to be updated to spki?

kevincennis: Sorry, I don’t understand the question

davidlehn: These resolved things are confusing me, i'll comment in the latest code.

lib/rsa.js
```diff
@@ -887,6 +889,28 @@ pki.rsa.generateKeyPair = function(bits, e, options, callback) {
   // key generation code if available and if parameters are acceptable
   if(!forge.options.usePureJavaScript && callback &&
     bits >= 256 && bits <= 16384 && (e === 0x10001 || e === 3)) {
+    if(_detectNodeCrypto('generateKeyPair')) {
+      return _crypto.generateKeyPair('rsa', {
+        modulusLength: bits,
+        publicExponent: e,
+        publicKeyEncoding: {
+          type: 'pkcs1',
```
davidlehn: Should this be 'spki'?

lib/util.js
```diff
@@ -125,6 +125,11 @@ util.isArrayBufferView = function(x) {
   return x && util.isArrayBuffer(x.buffer) && x.byteLength !== undefined;
 };
 
+// Node versions >= 10.12.0 support native keyPair generation,
+// which is non-deterministic
+util.isDeterministic = util.isNodejs &&
+  typeof require('crypto').generateKeyPair !== 'function';
```
davidlehn: It seems odd to export this as part of the main API.  Maybe the name is just too generic?  Is this something people would use outside of testing?  If not it could just be in the rsa test file.

lib/rsa.js
kevincennis: Seems like `pki.publicKeyFromPem` will take either. I used `pkcs1` because that's what was done with `window.crypto.subtle.generateKey` a few lines below, but I can change to `spki` if you think that makes more sense.

lib/util.js
kevincennis: Yeah, that's fair. Will move to the test file.

lib/rsa.js
davidlehn: @dlongley Do you have an opinion on this?  I'm not sure what the ramifications are one way or the other.

davidlehn:
@kevincennis I made some modifications on this PR in https://github.com/digitalbazaar/forge/pull/643.  Take a look and see if it's good.

davidlehn:
Merged via https://github.com/digitalbazaar/forge/pull/643.

PR #641, opened 2019-01-15
README: Remove circular documentation reference

jgarzik:
Main README references testing README, which is a one-line file
pointing the reader back to main.


davidlehn:
Also see #641.

PR #651, opened 2019-02-08
Forge 650

### LINKED ISSUE #650: x509 binary extracted from PKCS12 differs slightly from actual X.509
I have recently stumbled across a certificate, that when reencoded by node-forge results in a certificate that differs slightly from the actual certificate.

MCV sample (forge 0.8.0): https://jsfiddle.net/9m5v4ct7/3/

The problem appears to be an AlgorithmIdentifier that is PSS with exotic but technically valid Parameters (unfortunately used in some productive certificates from a large CA):

```
 SEQUENCE
  OBJECT            :rsassaPss
  SEQUENCE
   cont [ 0 ]
    SEQUENCE
     OBJECT            :sha256
   cont [ 1 ]
    SEQUENCE
     OBJECT            :mgf1
     SEQUENCE
      OBJECT            :sha256
```

When encoding the certificate extracted from the PKCS#12, forge emits the following AlgorithmIdentifier in the unsigned parts of the certificate:

```
 SEQUENCE
  OBJECT            :rsassaPss
  SEQUENCE
   cont [ 0 ]
    SEQUENCE
     OBJECT            :sha256
     NULL
   cont [ 1 ]
    SEQUENCE
     OBJECT            :mgf1
     SEQUENCE
      OBJECT            :sha256
      NULL
   cont [ 2 ]
    INTEGER           :14
```

This is semantically equivalent, but not binary identical to the AlgId inside the TBSCertificate. There are actually two different issues here:

1. the usual NULL parameter is added to the digest algorithm identifiers, although not originally present
2. the saltLength 20 (0x14) is emitted, but should not be included since it is the default value (although it should be 32 (0x20) when used with SHA-256, but technically ...)

I will work on a patch that addresses these issues. Is there something I should look out for?

MaxFichtelmann:
fix #650 by adding an additional property `signatureAlgorithm` onto the cert object created by `pki.certificateFromAsn1`.

when encoding the certificate and the above `signatureAlgorithm` property is present, it is used instead of constructing the algorithm from oid and parameters. In its absence the original approach is used, so this should only have an impact when encoding a previously decoded certificate.

### DIFF
--- a/lib/x509.js
+++ b/lib/x509.js
@@ -298,6 +298,7 @@ var x509CertificateValidator = {
     tagClass: asn1.Class.UNIVERSAL,
     type: asn1.Type.SEQUENCE,
     constructed: true,
+    captureAsn1: 'certSignatureAlgorithm',
     value: [{
       // algorithm
       name: 'Certificate.signatureAlgorithm.algorithm',
@@ -1291,6 +1292,7 @@ pki.certificateFromAsn1 = function(obj, computeHash) {
     capture.certVersion.charCodeAt(0) : 0;
   var serial = forge.util.createBuffer(capture.certSerialNumber);
   cert.serialNumber = serial.toHex();
+  cert.signatureAlgorithm = capture.certSignatureAlgorithm;
   cert.signatureOid = forge.asn1.derToOid(capture.certSignatureOid);
   cert.signatureParameters = _readSignatureParameters(
     cert.signatureOid, capture.certSignatureParams, true);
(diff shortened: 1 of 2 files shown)


MaxFichtelmann:
could you possibly take a look, if this solution fits? Is there a problem that prevents it to be merged?

davidlehn:
Sorry, we've been busy on other projects and haven't had a chance to get to this.  At some point we'll find the time for a proper review.  Two small things I noticed that should be changed:
- Files in that `test/issues/` dir are not run automatically.  I assume your new tests should always run so they should be added to the files in `tests/unit/`.  Probably tacked onto `x509.js`?  Could just use a comment to say what issue they are addressing.
- Use 2 space indents in the test code to match the other code.

MaxFichtelmann:
thank you, will do that.

Issue #657, opened 2019-02-22
Possible (though unprobable to hit) bug in kem.js


maltemuth:
I'm not knowledgeable enough in crypto stuff to follow the code and discuss it, either way I believe that the following line should also checkout that `r > 1` instead of just `r > 0`:

https://github.com/digitalbazaar/forge/blob/dfdde475677a8a25c851e33e8f81dca60d90cfb9/lib/kem.js#L63

Could someone knowledgeable enough kindly check this?

dlongley:
> Could someone knowledgeable enough kindly check this?

Yes, looks like a bug. PR #658 should fix it.

davidlehn:
Fixed in 0.8.1.  Thanks!

PR #658, opened 2019-02-23
Fix kem off-by-1 bug w/random generation.

dlongley:
Addresses #657.

### DIFF
--- a/lib/kem.js
+++ b/lib/kem.js
@@ -53,14 +53,14 @@ forge.kem.rsa.create = function(kdf, options) {
    *   key: the secret key to use for encrypting a message.
    */
   kem.encrypt = function(publicKey, keyLength) {
-    // generate a random r where 1 > r > n
+    // generate a random r where 1 < r < n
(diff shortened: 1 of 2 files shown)


lib/kem.js
```diff
@@ -53,14 +53,14 @@ forge.kem.rsa.create = function(kdf, options) {
...
-    // generate a random r where 1 > r > n
+    // generate a random r where 1 < r < n
     var byteLength = Math.ceil(publicKey.n.bitLength() / 8);
     var r;
     do {
       r = new BigInteger(
         forge.util.bytesToHex(prng.getBytesSync(byteLength)),
         16).mod(publicKey.n);
-    } while(r.equals(BigInteger.ZERO));
+    } while(r.compareTo(BigInteger.ONE) <= 0);
```
davidlehn: https://github.com/digitalbazaar/forge/blob/master/lib/jsbn.js#L259
Should this be `> 0`?

davidlehn: Hmm.  I think my logic is backwards.  Nevermind.

PR #661, opened 2019-03-06
PKCS#7: Custom message digest

vintagepx:
### Problem

If one is doing an detached signing, buffering the whole data to sign it is really inconvenient.

### Solution

My approach here is to allow a user-definied message digest, so you can code something like this: 

```js
const hash = crypto.createHash('sha512');
  
stream.on('data', chunk => hash.update(chunk));
stream.on('end', () => {
  const messageDigest = forge.util.createBuffer(hash.digest().toString('binary'));
  const p7 = pkcs7.createSignedData();
  
  p7.addSigner({
    key: privateKey,
    certificate: certificate,
    digestAlgorithm: forge.pki.oids.sha512,
    authenticatedAttributes: [{
      type: forge.pki.oids.contentType,
      value: forge.pki.oids.data
    },{
      type: forge.pki.oids.signingTime,
      value: new Date()
    },{
      type: forge.pki.oids.messageDigest,
      value: messageDigest
    }]
  });
  p7.sign({ detached: true });
});
```

Maybe a better way to solve this is by allowing `p7.content` to be able to receive a data stream, but for now this seems to solve the problem quite well.

### DIFF
--- a/lib/pkcs7.js
+++ b/lib/pkcs7.js
@@ -444,43 +444,41 @@ p7.createSignedData = function() {
   }
 
   function addSignerInfos(mds) {
-    var content;
+    var content,
+        bytes;
 
-    if (msg.detachedContent) {
-      // Signature has been made in detached mode.
-      content = msg.detachedContent;
-    } else {
-      // Note: ContentInfo is a SEQUENCE with 2 values, second value is
-      // the content field and is optional for a ContentInfo but required here
-      // since signers are present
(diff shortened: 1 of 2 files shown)


Simolation:
Is it possible to merge this PR, as this would be really useful for my use case?

nanndoj:
@davidlehn Any drawback to this approach? I am glad to help this PR to get merged.

PR #675, opened 2019-04-30
Add X.509 extensions

pmerlo:


pmerlo:
Hi,
I added the missing X.509 extensions `subjectKeyIdentifier` and `authorityKeyIdentifier` with all flags enabled. Please let me know if this should be changed and I'll do it.
Cheers!
Pedro

PR #678, opened 2019-05-20
Deprecated node buffer

aljones15:
removes calls on node's new Buffer. This gets rid of deprecation notices.

> (node:45127) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Forge ChangeLog
 
 ## 0.8.4 -
 
-### Removed
+### Changed
 - Replace all instances of Node.js `new Buffer` with `Buffer.from` and `Buffer.alloc`.
 
 ## 0.8.3 - 2019-05-15
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Forge ChangeLog
 ## 0.8.4 -
 
 ### Removed
-- All instances of Node.js `new Buffer` in favor of `Buffer.from` and `Buffer.alloc`.
+- Replace all instances of Node.js `new Buffer` with `Buffer.from` and `Buffer.alloc`.
 
 ## 0.8.3 - 2019-05-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Forge ChangeLog
 ## 0.8.4 -
 
 ### Removed
-- All instances of node new Buffer in favor of Buffer.from and Buffer.alloc
+- All instances of Node.js `new Buffer` in favor of `Buffer.from` and `Buffer.alloc`.
 
 ## 0.8.3 - 2019-05-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Forge ChangeLog
 ===============
 
+
+## 0.8.4 -
+
+### Removed
+- All instances of node new Buffer in favor of Buffer.from and Buffer.alloc
+
 ## 0.8.3 - 2019-05-15
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -1965,7 +1965,7 @@ var nodeBuffer = Buffer.from(forgeBuffer.getBytes(), 'binary');
 
 // convert a Node.js Buffer into a forge buffer
 // make sure you specify the encoding as 'binary'
-var nodeBuffer = Buffer.alloc(10);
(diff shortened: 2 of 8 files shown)


mattcollier:
I note that we're still building forge in CI for node 4.x: https://travis-ci.org/digitalbazaar/forge/builds/532979766?utm_source=github_status&utm_medium=notification

Docs state that `Buffer.from` was introduced in node 5.10: https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_array

Is it time to drop Node 4.x?

aljones15:
good catch not sure.

mattcollier:
Anyone know how/why CI is passing for 4.x *with* this patch?

dlongley:
@mattcollier,

It was added in 4.5.x:

https://nodejs.org/docs/latest-v4.x/api/buffer.html#buffer_class_method_buffer_from_str_encoding

mattcollier:
OK, thanks for nothing Node documentation.  Does that mean

> @mattcollier,
> 
> It was added in 4.5.x:
> 
> https://nodejs.org/docs/latest-v4.x/api/buffer.html#buffer_class_method_buffer_from_str_encoding

OK, thanks for nothing Node documentation that I found.  Does that mean this is good to go?

README.md
```diff
@@ -1961,11 +1961,11 @@ bytes.getBytes(/* count */);
 // convert a forge buffer into a Node.js Buffer
 // make sure you specify the encoding as 'binary'
 var forgeBuffer = forge.util.createBuffer();
-var nodeBuffer = new Buffer(forgeBuffer.getBytes(), 'binary');
+var nodeBuffer = Buffer.from(forgeBuffer.getBytes(), 'binary');
 
 // convert a Node.js Buffer into a forge buffer
 // make sure you specify the encoding as 'binary'
-var nodeBuffer = new Buffer();
+var nodeBuffer = Buffer.alloc(10);
```
davidlehn: `10` is arbitrary and not easy to understand as an example, let's either adjust this to something that makes more sense, or just remove it.
```suggestion
```

aljones15: what would make more sense?

MaxFichtelmann: Maybe a sample buffer with some content?

```suggestion
- var nodeBuffer = Buffer.alloc(10);
+ var nodeBuffer = Buffer.from('CAFE', 'hex');
```

davidlehn:
The package.json engines field should also be updated to ">= 4.5.0" (or whatever version we can figure out is appropriate).

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 Forge ChangeLog
 ===============
 
+
+## 0.8.4 -
+
+### Removed
+- All instances of node new Buffer in favor of Buffer.from and Buffer.alloc
```
davidlehn:
```suggestion
- All instances of Node.js `new Buffer` in favor of `Buffer.from` and `Buffer.alloc`.
```

aljones15:
hey used nvm to set the node to 4.5 all passing anything less and it fails. so set

```js
engine: {
  node: >= 4.5.0
}
```

aljones15:
@dlongley added a util function that calls new Buffer when less than 4.5 and Buffer.from when greater than 4.5.

all tests are passing in node 4.1 and node 4.5

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 Forge ChangeLog
 ===============
 
+
+## 0.8.4 -
+
+### Removed
+- All instances of Node.js `new Buffer` in favor of `Buffer.from` and `Buffer.alloc`.
```
dlongley:
```suggestion
- Replace all instances of Node.js `new Buffer` with `Buffer.from` and `Buffer.alloc`.
```

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 Forge ChangeLog
 ===============
 
+
+## 0.8.4 -
+
+### Removed
```
dlongley:
```suggestion
### Changed
```

aljones15:
This should be failing in karma because Buffer.from does not exist in the browser, but it is somehow passing...

davidlehn:
@aljones15 Looks like the Buffer code is all protected from running in a browser.  And flags are set to avoid webpack from adding in a polyfill.

PR #680, opened 2019-05-21
Asn1 validator node ed25519

aljones15:
Adds:

- ASN1 validators for der format ed25519 public and private keys
- Adds the OID for ed25519 keys
- Adds extract functions so node 12 ed25519 keys can be imported now.


dlongley:
Some of the docs in here make it seem like the functions are specific to node.js buffers, etc. -- when they aren't. The docs can mention that if you're using node's API that the functions will be particularly helpful, but they should lead by saying that the functions, for example, convert a DER-formatted PKCS#8 private key to a raw private key.

We may also want the functions to be something like `publicKeyFromDer()` and `publicKeyToDer()` ... or something that more closely approximates existing the forge API. I think we need a short bikeshedding around. cc: @mattcollier, @davidlehn

lib/ed25519.js
```diff
@@ -13,12 +13,21 @@ require('./jsbn');
 require('./random');
 require('./sha512');
 require('./util');
+require('./asn1');
```
davidlehn: Move up to be in alpha order.

lib/ed25519.js
```diff
@@ -36,6 +45,157 @@ ed25519.constants.SEED_BYTE_LENGTH = 32;
 ed25519.constants.SIGN_BYTE_LENGTH = 64;
 ed25519.constants.HASH_BYTE_LENGTH = 64;
 
+/**
+ * ASN1 public key validator for importing ed25519 keys from node >= 12
```
davidlehn:
```suggestion
 * ASN1 public key validator for importing ed25519 keys from Node.js >= 12
```

davidlehn: Also change "node" elsewhere in this PR to "Node.js" for clarity.

lib/ed25519.js
```diff
@@ -36,6 +45,157 @@ ed25519.constants.SEED_BYTE_LENGTH = 32;
...
+      ' recieved ' + oid);
+  }
+  const binary = capture[options.keyName];
+  return binary;
+};
+
+/**
+ * Takes in a key exported in DER and returns the raw key value.
+ *
+ * @param {NativeBuffer} publicKeyBuffer - A public key in der.
```
davidlehn:
```suggestion
 * @param {NativeBuffer} publicKeyBuffer - A public key in DER format.
```

tests/unit/ed25519.js
```diff
@@ -329,6 +337,35 @@ var UTIL = require('../../lib/util');
       ASSERT.equal(hex(signature), expectedSignature);
       ASSERT.equal(verified, true);
     });
+
+    it('should extract the key value from a der public key', function() {
```
davidlehn:
```suggestion
    it('should extract the key value from a DER public key', function() {
```

davidlehn: Also same change below.

lib/ed25519.js
aljones15: we're actually removing node in favor of DER formatted so other crypto libraries are included.

lib/ed25519.js
```diff
@@ -148,7 +148,7 @@ ed25519.extractKey = function(options) {
     throw new Error('Invalid Key: ' + errors.join('.'));
   }
   const oid = forge.asn1.derToOid(capture['ed25519Oid']);
-  const ed25519Oid = forge.oids['ed25519Encryption'];
+  const ed25519Oid = forge.oids['EdDSA25519'];
```
aljones15: thanks for adding this.

aljones15:
@mattcollier  sorry I had to change those consts but thanks for updating this.

lib/ed25519.js
```diff
@@ -36,6 +44,165 @@ ed25519.constants.SEED_BYTE_LENGTH = 32;
...
+ * @param {NativeBuffer} options.keyBuffer -
+ * The result of exporting the DER key.
+ * @param {ed25519PrivateKeyValidator|ed25519PublicKeyValidator}
+ * options.validator - A validator to check the exported key.
+ * @param {Object} [options.capture = {}] - A collection of name value pairs
+ * captured by the validator.
+ * @param {'ed25519PrivateKey'|'ed25519PublicKey'} options.keyName -
+ * The name of the key captured by the validator.
+ *
+ * @returns {string} - A binary string or the key value.
```
dlongley:
```suggestion
 * @returns {string} - A binary string of the key value.
```

lib/ed25519.js
```diff
@@ -36,6 +44,165 @@ ed25519.constants.SEED_BYTE_LENGTH = 32;
...
+  var errors = [];
+  var valid = forge.asn1.validate(asn1, options.validator, capture, errors);
+  if(!valid) {
+    throw new Error('Invalid Key: ' + errors.join('.'));
+  }
+  var oid = forge.asn1.derToOid(capture.ed25519Oid);
+  var ed25519Oid = forge.oids.EdDSA25519;
+  if(oid !== ed25519Oid) {
+    throw new Error('Invalid OID expected ' + ed25519Oid +
+      ' recieved ' + oid);
```
dlongley: Typo here on "recieved" but let's rewrite this error message to be more in line with others:

`'Invalid OID "' + oid + '"; OID must be "' + ed25519Oid + '".'`;

lib/ed25519.js
```diff
@@ -36,6 +44,165 @@ ed25519.constants.SEED_BYTE_LENGTH = 32;
...
+ * @param {ed25519PrivateKeyValidator|ed25519PublicKeyValidator}
+ * options.validator - A validator to check the exported key.
+ * @param {Object} [options.capture = {}] - A collection of name value pairs
+ * captured by the validator.
+ * @param {'ed25519PrivateKey'|'ed25519PublicKey'} options.keyName -
+ * The name of the key captured by the validator.
+ *
+ * @returns {string} - A binary string or the key value.
+ */
+ed25519.extractKey = function(options) {
```
dlongley: Let's not expose this method publicly yet, it will enable us to get this PR in sooner. There's a question about what the name and shape of this API should be or whether it should be public at all.

lib/ed25519.js
```diff
@@ -9,16 +9,24 @@
  * https://github.com/dchest/tweetnacl-js
  */
 var forge = require('./forge');
+require('./asn1');
```
dlongley: Not a huge fan of requiring `asn1` to use ed25519 ... but I think it will take too long to get this merged if we try and engineer around it. We can save that for another day.

lib/ed25519.js
```diff
@@ -36,6 +44,165 @@ ed25519.constants.SEED_BYTE_LENGTH = 32;
...
+ *
+ * @returns {string} - A binary string or the key value.
+ */
+ed25519.extractKey = function(options) {
+  var capture = options.capture || {};
+  var asn1 = getKeyBytes(options.keyBuffer);
+  var errors = [];
+  var valid = forge.asn1.validate(asn1, options.validator, capture, errors);
+  if(!valid) {
+    throw new Error('Invalid Key: ' + errors.join('.'));
```
dlongley: This is going to produce a pretty strange error message when it happens. Do we do this elsewhere? Let's copy what we do in another place -- whatever that may be.

```suggestion
    throw new Error('Invalid key; ' + errors.join('.'));
```

mattcollier:
@dlongley @aljones15 I'll handle Longley's comments.

lib/ed25519.js
dlongley: Here's an example to copy from RSA:

```js
    var error = new Error('Cannot read private key. ' +
      'ASN.1 object does not contain an RSAPrivateKey.');
    error.errors = errors;
```

lib/ed25519.js
```diff
@@ -36,6 +44,132 @@ ed25519.constants.SEED_BYTE_LENGTH = 32;
...
+    type: asn1.Type.BITSTRING,
+    constructed: false,
+    composed: true,
+    captureBitStringValue: 'ed25519PublicKey'
+  }]
+};
+
+/**
+ * ASN1 private key validator for importing ed25519 keys in DER format.
+ * @type {Object}
```
dlongley: Please add a note here:

```
This validator is for a `OneAsymmetricKey` ASN.1 type as specified in RFC5958 (https://tools.ietf.org/html/rfc5958). This type has the same format as a simple `PrivateKeyInfo`
for which there is already a validator specified in forge's `rsa` module.

TODO: Reuse `PrivateKeyInfo` validator from forge `rsa`.
```

lib/ed25519.js
```diff
@@ -36,6 +44,132 @@ ed25519.constants.SEED_BYTE_LENGTH = 32;
...
+  tagClass: asn1.Class.UNIVERSAL,
+  type: asn1.Type.SEQUENCE,
+  constructed: true,
+  composed: true,
+  value: [{
+    tagClass: asn1.Class.UNIVERSAL,
+    type: asn1.Type.INTEGER,
+    constructed: false,
+    composed: false,
+    capture: 'privateKeyZero' // This value should always be 0.
```
dlongley:
```suggestion
    capture: 'privateKeyVersion' // only version 1 (which is value `0`) is supported
```

lib/ed25519.js
```diff
@@ -36,6 +44,132 @@ ed25519.constants.SEED_BYTE_LENGTH = 32;
...
+    composed: true,
+    captureBitStringValue: 'ed25519PublicKey'
+  }]
+};
+
+/**
+ * ASN1 private key validator for importing ed25519 keys in DER format.
+ * @type {Object}
+*/
+var ed25519PrivateKeyValidator = forge.ed25519.privateKeyValidator = {
```
dlongley:
```suggestion
var ed25519PrivateKeyValidator = {
```

dlongley: Removing this from public exposure because we should be reusing other things.

lib/ed25519.js
```diff
@@ -36,6 +44,132 @@ ed25519.constants.SEED_BYTE_LENGTH = 32;
 ed25519.constants.SIGN_BYTE_LENGTH = 64;
 ed25519.constants.HASH_BYTE_LENGTH = 64;
 
+/**
+ * ASN1 public key validator for importing ed25519 keys in DER format.
+ * @type {Object}
+*/
+var ed25519PublicKeyValidator = forge.ed25519.publicKeyValidator = {
```
dlongley:
```suggestion
var ed25519PublicKeyValidator = {
```

lib/ed25519.js
dlongley: If helpful, we can include the link to the other validator, but it could get out of date.

https://github.com/digitalbazaar/forge/blob/master/lib/rsa.js#L96

mattcollier:
@davidlehn I leave it to you to approve/merge and publish this.  I'm not an owner on NPM.  Thanks!

lib/ed25519.js
```diff
@@ -36,6 +44,140 @@ ed25519.constants.SEED_BYTE_LENGTH = 32;
...
+ */
+function getKeyBytes(keyBuffer) {
+  var binary = forge.util.binary.raw.encode(keyBuffer);
+  return forge.asn1.fromDer(binary);
+}
+
+/**
+ * Takes in a public key exported in DER and returns the raw key value.
+ *
+ * @param {NativeBuffer} publicKeyBuffer - A DER formatted public key.
```
dlongley: Any use of `NativeBuffer` in the docs should change -- this isn't a public name. It should be `{Uint8Array|Buffer}`.

aljones15: This actually is a jsdoc thing NativeBuffer here refers to a jsdoc type declared earlier in the file:

```
/**
 * NativeBuffer is either a Buffer or a Uint8Array.
 * @type {Buffer|Uint8Array}
*/
```

aljones15:
@dlongley @mattcollier this can be closed correct? another PR added this feature I believe.

dlongley:
I don't know what the status is, but so we can better track it I'm linking to that other PR: #697.

davidlehn:
#697 was merged and released causing conflicts here.  If some of this patch should still be merged, please rebase and update as needed.

aljones15:
if I recall correctly this PR is not needed because it turns out the ed25519 keys use a common DER format that we already have a validator for. I believe this can be closed @dlongley

PR #695, opened 2019-07-16
Add oid for EdDSA25519.

mattcollier:
OID data taken from: http://oid-info.com/get/1.3.101.112 and 

It's not clear to me if the id should be `Ed25519`
https://tools.ietf.org/id/draft-ietf-curdle-pkix-01.html#rfc.section.3

or `EdDSA25519`
https://tools.ietf.org/id/draft-ietf-curdle-pkix-01.html#rfc.section.5

### DIFF
--- a/lib/oids.js
+++ b/lib/oids.js
@@ -34,6 +34,8 @@ _IN('1.2.840.113549.1.1.10', 'RSASSA-PSS');
 _IN('1.2.840.113549.1.1.11', 'sha256WithRSAEncryption');
 _IN('1.2.840.113549.1.1.12', 'sha384WithRSAEncryption');
 _IN('1.2.840.113549.1.1.13', 'sha512WithRSAEncryption');
+// Edwards-curve Digital Signature Algorithm (EdDSA) Ed25519
+_IN('1.3.101.112', 'EdDSA25519');
 
 _IN('1.2.840.10040.4.3', 'dsa-with-sha1');
 


aljones15:
this is actually duplicate work:

https://github.com/digitalbazaar/forge/pull/680/files

That Pr already contains the oid, but no problem if we need a smaller Pr while whatever everyone is to busy to the review the other one.

mattcollier:
> That Pr already contains the oid, but no problem if we need a smaller Pr while whatever everyone is to busy to the review the other one.

Sorry, I apparently looked at your PR before the commit that put the OID in the file.  I see that you've labelled it as an encryption OID which I'm pretty sure is not accurate.

Also regarding your other PR, I suggest that moving the linting corrections into another PR may make it easier to approve.  I would like to see that PR get merged because the work I'm doing in crypto-ld relies on it as well.

davidlehn:
I'm no expert on these ed/dsa/x 25519 issues and naming.  Those draft links are broken and looks like it got to an RFC: https://tools.ietf.org/html/rfc8410.

aljones15:
> > That Pr already contains the oid, but no problem if we need a smaller Pr while whatever everyone is to busy to the review the other one.
> 
> Sorry, I apparently looked at your PR before the commit that put the OID in the file. I see that you've labelled it as an encryption OID which I'm pretty sure is not accurate.
> 
> Also regarding your other PR, I suggest that moving the linting corrections into another PR may make it easier to approve. I would like to see that PR get merged because the work I'm doing in crypto-ld relies on it as well.

I removed the 2 lint only commits unfortunately there was a correction in the ed25519 file. I can use your oid once your PR is merged.

aljones15:
btw regarding the naming issue why just include both? then we account for what happens if one user used ed25519 and one used edDSA25519?

PR #697, opened 2019-07-22
Implement ed25519.publicKeyFromAsn1.

mattcollier:
Strawman WIP, @dlongley please provide feedback.

Is the return value for `publicKeyFromAsn1`, a NativeBuffer with the key bytes correct?

### DIFF
--- a/lib/ed25519.js
+++ b/lib/ed25519.js
@@ -161,9 +161,13 @@ ed25519.sign = function(options) {
     message: options.privateKey,
     encoding: 'binary'
   });
-  if(privateKey.length !== ed25519.constants.PRIVATE_KEY_BYTE_LENGTH) {
+  if(privateKey.length === ed25519.constants.SEED_BYTE_LENGTH) {
+    var keyPair = ed25519.generateKeyPair({seed: privateKey});
+    privateKey = keyPair.privateKey;
+  } else if(privateKey.length !== ed25519.constants.PRIVATE_KEY_BYTE_LENGTH) {
     throw new TypeError(
       '"options.privateKey" must have a byte length of ' +
+      ed25519.constants.SEED_BYTE_LENGTH + ' or ' +
       ed25519.constants.PRIVATE_KEY_BYTE_LENGTH);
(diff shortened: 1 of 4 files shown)


aljones15:
looks nice.

mattcollier:
See how these APIs work in practice: https://github.com/digitalbazaar/crypto-ld/blob/node-12/lib/Ed25519KeyPair.js#L80-L143

Is this what we're looking for?

Note that the implementation in crypto-ld is expecting just 32 bytes from the private key decoder.

mattcollier:
Tests are passing using this in crypto-ld: https://github.com/digitalbazaar/crypto-ld/pull/34

The fixme's in the code are concerned with ensuring that the return values for the functions are what we want which is 32 key bytes in both public/private cases.  Some APIs consider the private key to be 64 bytes where the public key is prepended to the private key.  Currently, this concatenation is happening in crypto-ld: https://github.com/digitalbazaar/crypto-ld/pull/34/files#diff-b4c6cf9ee887b894feccc1931bb592afR105-R109

lib/ed25519.js
```diff
@@ -64,6 +67,67 @@ ed25519.generateKeyPair = function(options) {
...
+  if(oid !== ed25519Oid) {
+    throw new Error('Invalid OID "' + oid + '"; OID must be "' +
+      ed25519Oid + '".');
+  }
+  var privateKey = capture.privateKey;
+  // manually extract the private key bytes from nested octet string, see FIXME:
+  // https://github.com/digitalbazaar/forge/blob/master/lib/asn1.js#L542
+  var privateKeyBytes = forge.asn1.fromDer(privateKey).value;
+  // FIXME: what should be returned from this API?  Should this compose the
+  // 64 byte public key?
```
dlongley: This should be consistent with how the keys are being used elsewhere in forge. It looks like it's 64 bytes from the above constant, so we should use that. It would be very unusual for people to use forge's `fromASN1` API and get a value that they can't use with the other APIs.

PR #704, opened 2019-08-21
Upgrade webpack version to 4.7.0

kannanvm:
Update Webpack to v4.x series.


davidlehn:
At some point a similar change was applied. Closing.
Of course now we need to move to v5.

Issue #721, opened 2019-09-26
[BUG] For block ciphers with block size <16, CTR mode doesn't mutate the counter


cbeuw:
For instance
```js
var forge = require('node-forge');

var key = forge.util.hexToBytes('0011223344556677')
var input = "The quick brown fox jumps over the lazy dog"
var iv = forge.util.hexToBytes('00')

var cipher = forge.cipher.createCipher('DES-CTR', key);
cipher.start({iv:iv});
cipher.update(forge.util.createBuffer(input));
cipher.finish();
console.log(cipher.output.toHex())
```
Yields `700abe5fad7509b94f42b90db3770efa420da35fb6750daa5742b409b97240ae4c07fb13bd7a19fa400dbc51`
However, both GoLang's standard library and pyCryptoDome gives `700abe5fad7509b9bd9e9a3514e4a6d9536d0c9b2e51e7a2a611db37b45dd99a0c2064464a77a377f791b11b`

The results deviate after `700abe5fad7509b9`, which is exactly one block
```
wrong: 700abe5fad7509b94f42b90db3770efa420da35fb6750daa5742b409b97240ae4c07fb13bd7a19fa400dbc51
right: 700abe5fad7509b9bd9e9a3514e4a6d9536d0c9b2e51e7a2a611db37b45dd99a0c2064464a77a377f791b11b
```

This is due to that:
- In cipherModes.js, IV is transformed into an array of four 32-bit ints (which is 16 bytes in total) regardless of the blocksize 
https://github.com/digitalbazaar/forge/blob/1af95997c4c555cf96e954a395c4295759f00fd7/lib/cipherModes.js#L447 https://github.com/digitalbazaar/forge/blob/1af95997c4c555cf96e954a395c4295759f00fd7/lib/cipherModes.js#L972-L973 `_inBlock` is then set to this transformed IV to be fed into the algorithm's block encrypt function.

  When IV is 8 bytes in size, as is the case for ciphers with block size of 8, the first two elements of the initial `_inBlock` will be the correct two int32s representing the entire IV, and the last two will be 0

- In des.js, only the first two int32 of the input block are taken into account. https://github.com/digitalbazaar/forge/blob/1af95997c4c555cf96e954a395c4295759f00fd7/lib/des.js#L364-L365 Since the first two elements do indeed represent the IV, the result of the first block is correct

[...]

PR #723, opened 2019-09-26
Address Issue #721: Increment CTR mode counter at the correct position when blocksize is not 16

cbeuw:
I altered `inc32` function so that instead of incrementing the last 32 bits with respect to the buffer's length, the caller needs to provide the index of the last 32 bits depending on the blocksize of the cipher

### DIFF
--- a/lib/cipherModes.js
+++ b/lib/cipherModes.js
@@ -501,7 +501,7 @@ modes.ctr.prototype.encrypt = function(input, output, finish) {
   }
 
   // block complete, increment counter (input block)
-  inc32(this._inBlock);
+  inc32(this._inBlock, this._ints - 1);
 };
 
 modes.ctr.prototype.decrypt = modes.ctr.prototype.encrypt;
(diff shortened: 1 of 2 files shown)


dlongley:
@cbeuw -- Thanks! I do prefer the other PR's fix as it deals with bad IV inputs and doesn't require the user of the utility function to know the details of `inc32`, but this is great that it has tests!

@davidlehn -- can you do some magic to incorporate the tests from this PR and the fix from #722?

dlongley:
@cbeuw, if you're able to help @davidlehn out by rebasing this PR on #722 with just the test commits, that would be most appreciated!

cbeuw:
Of course! Although this may be beyond my limited git skills :P 

I created a new branch and cherry-picked the two test commits onto that. Should I now make a pull request to merge this onto `fix-721`?

davidlehn:
I merged testing bits from this PR and the code fix from #722 into master.  I think this can be closed.

PR #726, opened 2019-10-13
Replace floating point operations with integer ops

paragonie-scott:
Context: https://twitter.com/CiPHPerCoder/status/1183435549442486272


dlongley:
The ed25519 internals all use Float64Arrays for consistency and to ensure there are a sufficient number of bits. For more discussion, see: https://github.com/dchest/tweetnacl-js/issues/45

PR #727, opened 2019-10-15
Accept CSRs with NEW in the label for decoding.

mundry:
This PR makes forge accept CSRs with `BEGIN NEW CERTIFICATE REQUEST` and `END NEW CERTIFICATE REQUEST` labels for decoding.

[1] https://tools.ietf.org/html/rfc7468#section-7

### DIFF
--- a/lib/pem.js
+++ b/lib/pem.js
@@ -96,7 +96,7 @@ pem.decode = function(str) {
   var rval = [];
 
   // split string into PEM messages (be lenient w/EOF on BEGIN line)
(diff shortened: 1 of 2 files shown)


davidlehn:
Could you a test for this?  Perhaps something in the x509 tests that makes sure the type is captured properly without "NEW".

davidlehn:
Thanks.  I adjusted how this works and merged it.  Decided to use a special case to convert "NEW CERTIFICATE REQUEST" to "CERTIFICATE REQUEST" rather than dropping "NEW" in the regex.  The regex could cause a problem in two cases:
- Accepting incorrect types.  Like "BEGIN NEW PUBLIC KEY" would be handled as "PUBLIC KEY".
- Unable to accept something else that should have NEW.  I'm not sure if that is a real thing.  But it would make accepting "BEGIN NEW FOO" as a type of "NEW FOO" impossible.

https://github.com/digitalbazaar/forge/pull/939

Issue #397, opened 2016-05-11
Please add support for UTF8 in x509 subject


Cojad:
https://github.com/digitalbazaar/forge/blob/master/js/x509.js#L1959

I found the code is default with encoding  with PRINTABLESTRING.
Which user have no option to modify the default encoding to utf8 or I missed?
node-forge would produce illegal cert when subject contains UTF8 characters.  
I have modify this line to UTF8 in order to produce my certs in correct encoding. I'm not familiar for node.js programing and it took me almost a day to find here to get my job done. But I hope there will be option for change default so one doesn't need to modify the code in node-forge.

Cojad:
I apologized. There is indeed a way to set utf8 for subjects.
```
var attr = [
        {name: 'countryName', value: 'TW' valueTagClass: 12},
        {name: 'organizationName', value: 'ass.tw', valueTagClass: 12},
        {shortName: 'ST', value: 'Chinese 大安區', valueTagClass: 12},
        {shortName: 'CN', value: 'Chinese 中文', valueTagClass: 12},
        {shortName: 'OU', value: 'Chinese 中文部門', valueTagClass: 12},
        {shortName: 'CN', value: 'Chinese 中文', valueTagClass: 12}
];
 ```
The key is set "valueTagClass" with a value 12 which indicate UTF8. Then you can put UTF8 character in value and produce well formed certs.

Cojad:
However this didn't fix all the problem. I soon realized that node-forge didn't decode x509 pem correctly with utf8 subject. So I fix it be add a few line in my fork. I'm afraid that my commit doesn't meet the contribute guild. So I din't create a poll from my commit. But feel free to take it.

https://github.com/Cojad/forge/commit/0f6d5d3c718666e871aaac1bae3f29374369f743

M-OC:
+1

genaromadrid:
Great @Cojad, your change also fixes UTF8 characters in the issuer... this was exactly what i has looking for

Cojad:
I'm glad that someone find my patch useful. :)

Djalmar:
Another solution I have found is to use the valueTagClass field on the subject, when the value is 12 you should decode to fix the problems.

GitStorageOne:
Still actual, right?

kamil7x:
@davidlehn @dlongley any chances that patch created by @Cojad will be picked up by you? Or is there any other solution? I still run into decoding problem and I would like not to fork library.

[...]

PR #731, opened 2019-11-05
Support for UTF-8 chars in subject

kamil7x:
Fixes https://github.com/digitalbazaar/forge/issues/397
Based on https://github.com/digitalbazaar/forge/issues/397#issuecomment-218420565

### DIFF
--- a/lib/x509.js
+++ b/lib/x509.js
@@ -520,7 +520,8 @@ pki.RDNAttributesAsArray = function(rdn, md) {
       obj.valueTagClass = attr.value[1].type;
       if(obj.valueTagClass === asn1.Type.UTF8) {
         obj.value = forge.util.decodeUtf8(obj.value);
-      }      // if the OID is known, get its name and short name
+      }      
+      // if the OID is known, get its name and short name
       if(obj.type in oids) {
         obj.name = oids[obj.type];
         if(obj.name in _shortNames) {
--- a/lib/x509.js
+++ b/lib/x509.js
@@ -518,7 +518,9 @@ pki.RDNAttributesAsArray = function(rdn, md) {
       obj.type = asn1.derToOid(attr.value[0].value);
       obj.value = attr.value[1].value;
       obj.valueTagClass = attr.value[1].type;
-      // if the OID is known, get its name and short name
+      if(obj.valueTagClass === asn1.Type.UTF8) {
+        obj.value = forge.util.decodeUtf8(obj.value);
+      }      // if the OID is known, get its name and short name
       if(obj.type in oids) {
         obj.name = oids[obj.type];
         if(obj.name in _shortNames) {


LuanMaik:
Why not merge?

davidlehn:
@LuanMaik This fell off the radar.  It's a challenge to determine if this is the correct fix.  Ideally this would refer to the proper part of a spec that defines the behavior, and there would be some tests added.  Data in the wild may or may not follow those specs, so that adds more complexity.

Does anyone have time to add test cases for this?

LuanMaik:
This is a relevant case to be observed
https://github.com/vbuch/node-signpdf/issues/111

bopos:
Hello,

I did a test with two V3 certifcates in the context of pdf signature

- first one with special chars **CN = AC Representación** 
- second one with **CN = AC Representacion** (without special chars)

Without the fix the first one fail
The code called in node-signpdf is https://github.com/vbuch/node-signpdf/blob/develop/src/signpdf.js#L80

```
    const forgeCert = _nodeForge.default.util.createBuffer(p12Buffer.toString('binary'));
    const p12Asn1 = _nodeForge.default.asn1.fromDer(forgeCert);
    const p12 = _nodeForge.default.pkcs12.pkcs12FromAsn1(p12Asn1, options.asn1StrictParsing, options.passphrase);
```
For me this PR fix my problem

SrZorro:
Hi @davidlehn, could this PR be merged? Or its missing something? The related issue #397 its marked as resolved, but the PR its not merged, there is something we can do to merge it?

At my company we are having the same problem as vbuch/node-signpdf#111 and we ended here, for now we are working with a local fork, but we would like to know if we can do anything to help moving this forward, thanks.

LuanMaik:
What do you think to change this:
```js
if(obj.valueTagClass === asn1.Type.UTF8) {
     obj.value = forge.util.decodeUtf8(obj.value);
} 
```

to this:
```js
if(obj.valueTagClass === asn1.Type.UTF8) {
   try {
       obj.value = forge.util.decodeUtf8(obj.value);
   } catch(e) { 
      /* This exception means obj.value is utf-8 */ 
   }
} 
```

to avoid unhandled exceptions, when the `obj.value` is utf-8 already.

Check: https://stackoverflow.com/questions/36314943/check-if-javascript-string-is-valid-utf-8

Arturokin:
I need this fix, why not merge?

LuanMaik:
@Arturokin It's necessary to add test cases for this to determine if this is the correct fix.

LuanMaik:
I give up and created the solution using Quarkus (java) and PDFBox lib.
This link show how to implement the solution in java: https://jvmfy.com/2018/11/17/how-to-digitally-sign-pdf-files/

kamil7x:
I am not using forge anymore. Maybe someone could jump in and fix this?

AgustinVillagran:
Hi @kamil7x, what are you using now to avoid this issue? I'm struggling with it.

PR #756, opened 2020-01-30
[BUG fix]Remove excess padding

fjh1997:
Remove excess padding when the length of input is just a multiple of the blocksize.
For example, when I input 128bit data with 128bit key in AES-CBC mode,the length of output should be 128bit,but the excess padding make it 256bit.
Obviously,it is a bug.

### DIFF
--- a/lib/cipherModes.js
+++ b/lib/cipherModes.js
@@ -70,8 +70,8 @@ modes.ecb.prototype.decrypt = function(input, output, finish) {
 modes.ecb.prototype.pad = function(input, options) {
   // add PKCS#7 padding to block (each pad byte is the
   // value of the number of pad bytes)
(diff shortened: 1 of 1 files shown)


dlongley:
Thanks for the PR! But, this is not a bug, it is intentional behavior. If your particular application *always* uses a fixed data size you can do your own custom padding, but the general behavior should not be modified.

See: https://crypto.stackexchange.com/questions/24130/why-can-the-last-block-contain-a-full-block-of-padding-in-cbc-encryption and https://crypto.stackexchange.com/questions/15058/what-prevents-a-padding-standard-to-cause-a-data-loss

Closing.

fjh1997:
thank you for explaination！

PR #764, opened 2020-02-23
Fix: avoid prime worker message spam.

dvogel:
The prime worker can receive messages posted by browser extensions (e.g. react-devtools-bridge). Since the message lacked the necessary inputs, the prime algorithm would start but never terminate. This drops useless messages on the floor.

### DIFF
--- a/lib/prime.worker.js
+++ b/lib/prime.worker.js
@@ -19,8 +19,10 @@ var BIG_TWO = new BigInteger(null);
 BIG_TWO.fromInt(2);
 
 self.addEventListener('message', function(e) {
-  var result = findPrime(e.data);
-  self.postMessage(result);
+  if (e.data.hex && e.data.workLoad) {
+    var result = findPrime(e.data);
+    self.postMessage(result);
+  }
 });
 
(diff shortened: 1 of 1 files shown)


davidlehn:
I'm not sure what to do with this.

On one hand, I can see how getting arbitrary messages could cause problems.  And I can see how this is a simple change that would fix it (other than in the edge case where the arbitrary message had the same fields).

On the other hand, I can't see why a browser extension would send arbitrary messages to a worker and *not* expect havoc to ensue.  What is being sent, and why?

Is there somewhere that suggests best practices for this?

dlongley:
I think it's fine to ensure that we're only acting on messages that are intended to be received by the worker. However, we should add a more specific message `type` property (with a value that is an identifier that specifies the message is a forge prime worker message) that we can check rather than relying on looking for `e.data.hex` and `e.data.workLoad`.

davidlehn:
Tried another approach in https://github.com/digitalbazaar/forge/pull/941.

dlongley:
Closing in favor of #941, which should be merged.

Issue #602, opened 2018-08-03
n.randomBytes is not a function


bradleyhodges:
When trying to execute `forge.pki.encryptRsaPrivateKey(keypair.privateKey, password);` in an Electron app, the exception `n.randomBytes is not a function` is thrown.

I've double checked, and this issue doesn't occur when executed in a Google Chrome tab. Here's the script I'm using:

```
var rsa = forge.pki.rsa; var password = "(this is retrieved elsewhere)";
rsa.generateKeyPair({bits: 2048, workers: 2}, function(err, keypair) {
    var publicKey = forge.pki.publicKeyToPem(keypair.publicKey);
    var privateKey = forge.pki.encryptRsaPrivateKey(keypair.privateKey, password);
});
```

By the looks of it, the exception is occurring when trying to execute the `forge.pki.encryptRsaPrivateKey` function in Electron.

Here's what the exception looks like in the DevTools:
![image](https://user-images.githubusercontent.com/15671567/43629418-c948a86c-972f-11e8-8902-8f6db91e3a66.png)

and this is where the exception appears to occur:
![image](https://user-images.githubusercontent.com/15671567/43629537-18dea4ee-9730-11e8-9cc4-cbed7fb00cf7.png)

KunalOjha:
@bradleyhodges has there been any updates or workarounds to this bug?

davidlehn:
Might be the node vs browser code path logic.  I'm not sure what happens in Electron.  Try running with non-minimized code so it's easier to see what it's doing.  Maybe that detection code can be fixed up.

mattnotmitt:
@davidlehn I'm currently having this issue, I've had a look in the minified JS and no `require('crypto')` call exists.

GitSage:
Easy workaround:

```
if (process && !process.versions['node-webkit']) {
  process.versions['node-webkit'] = true;
}
const forge = require('node-forge');
```

I'm looking into making a PR to handle Electron the same way as Node-Webkit.

GitSage:
PR submitted https://github.com/digitalbazaar/forge/pull/782

Thulof:
I have created a new PR https://github.com/digitalbazaar/forge/pull/1110 to fix it.

PR #782, opened 2020-05-26
Fix for issue #602. Chrome's window.crypto property is read-only, so …

GitSage:
…when Electron tries to require Node's crypto library, it fails. The fix is identical to that for node-webkit.

### DIFF
--- a/lib/prng.js
+++ b/lib/prng.js
@@ -14,7 +14,7 @@ require('./util');
 
 var _crypto = null;
 if(forge.util.isNodejs && !forge.options.usePureJavaScript &&
-  !process.versions['node-webkit']) {
+  !process.versions['node-webkit'] && !process.versions['electron']) {
   _crypto = require('crypto');
 }
 
--- a/lib/prng.js
+++ b/lib/prng.js
@@ -14,7 +14,7 @@ require('./util');
 
 var _crypto = null;
 if(forge.util.isNodejs && !forge.options.usePureJavaScript &&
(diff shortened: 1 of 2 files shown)


mattcollier:
Thank you for the PR.  I think this logic probably belongs in the `isNodejs` utility here:

https://github.com/digitalbazaar/forge/blob/master/lib/util.js#L110-L111

GitSage:
@mattcollier Sounds fine to me. PR updated.

mattcollier:
@GitSage it's my understanding that Electron provides two contexts where you main run code.  The main process and a renderer process.  The main process can load the node crypto module according to [this comment](https://github.com/electron/node/issues/92#issuecomment-462064813).

[This article](https://cameronnokes.com/blog/deep-dive-into-electron%27s-main-and-renderer-processes/) claims the full node API is available in the main process.

I suspect the difficulty here is in the renderer process which does not include the full Node API?

Does forge function properly in one execution context and not in the other?  If so, does this patch properly discern between the two?

GitSage:
@mattcollier I have investigated and you are correct, the main process can load the node crypto module correctly. The renderer process cannot.

The process of moving node-forge into the main thread was a pretty big pain. You have to perform asynchronous communication with the main thread through ICP, which carries a performance hit with it.

I think this PR should be closed without merging. Those who need this behavior should use the existing `usePureJavaScript` flag.

mattcollier:
@GitSage thanks for the follow-up.  Closing.

PR #809, opened 2020-09-01
Add extra condition to setPath function

tamascsaba:
It was possible to make prototype pollution  throught setPath function.
Now the code will exit when merging objects with sensitive properties,
such as constructor or __proto__.

See details here: 
https://snyk.io/vuln/SNYK-JS-NODEFORGE-598677

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -2531,6 +2531,9 @@ util.setPath = function(object, keys, value) {
     var len = keys.length;
     while(i < len) {
       var next = keys[i++];
+      if(next === '__proto__' || next === 'constructor' || next === 'prototype') {
+        return object;
+      }
       if(i == len) {
         // last
         object[next] = value;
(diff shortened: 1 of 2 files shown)


jaredkstraub:
+1 on this

davidlehn:
Thanks for the patch.  We're taking an alternate approach.  It's debatable if there was a security issue here or just a feature you have to use properly.  The CVEs forced the issue of dealing with it.  Any change, like what is proposed here, would be a breaking behavior change.  And since that would require a new major version release, we just decided to remove `getPath`, `setPath`, and `deletePath`.  There are better replacements in lodash.  I released a simple doc update in 0.9.2 to mention the issue.  0.10.0 removes the path functions.

If we were to try and update the functions, what you proposed is close but could probably use some changes.  I'd think those blocked properties should throw an error rather than just return.  The issue being that some parts of the path might have already been created and you might want to know if the call worked.  It would be confusing for the caller as is.  Probably better to scan the path keys first and fail if any are in a blocked list.  But that is all moot now, functions removed in favor of lodash or similar.  Thanks!

tamascsaba:
Thank you for your upgrade pls fix the issue in synk: https://support.snyk.io/hc/en-us/articles/360003891038-Fix-your-vulnerabilities

PR #823, opened 2020-09-25
fix: globalScope is undefined when running in JSCore Environment

smalldice:
Hello,
    An issue came when I use node-forge in JSCore environment,  at lib/random.js line 119, globalScope returns undefined, cause globalScope.crypto throw Error "can't read property crypto of undefined", so I changed lib/util.js getScopeFunction to fix this problem

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -120,7 +120,7 @@ util.globalScope = (function() {
     return global;
   }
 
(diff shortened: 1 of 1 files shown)


ProgrammerHuang:
The same problem at prng.js line 271.

PR #824, opened 2020-10-07
Feature/add unauthenticated attributes

Strnadj:
Right now is impossible to add unauthenticated attributes via the `addSigner` method. The method sets them to an empty array and ignores them later. (You can't add response from TSA for example).

### DIFF
--- a/lib/oids.js
+++ b/lib/oids.js
@@ -68,10 +68,10 @@ _IN('1.2.840.113549.1.9.7', 'challengePassword');
 _IN('1.2.840.113549.1.9.8', 'unstructuredAddress');
 _IN('1.2.840.113549.1.9.14', 'extensionRequest');
 
+_IN('1.2.840.113549.1.9.16.2.14', 'timeStampToken');
 _IN('1.2.840.113549.1.9.20', 'friendlyName');
 _IN('1.2.840.113549.1.9.21', 'localKeyId');
 _IN('1.2.840.113549.1.9.22.1', 'x509Certificate');
-_IN('1.2.840.113549.1.9.16.2.14', 'timeStampToken');
 
 // pkcs#12 safe bags
 _IN('1.2.840.113549.1.12.10.1.1', 'keyBag');
--- a/lib/oids.js
+++ b/lib/oids.js
@@ -71,6 +71,7 @@ _IN('1.2.840.113549.1.9.14', 'extensionRequest');
 _IN('1.2.840.113549.1.9.20', 'friendlyName');
 _IN('1.2.840.113549.1.9.21', 'localKeyId');
 _IN('1.2.840.113549.1.9.22.1', 'x509Certificate');
+_IN('1.2.840.113549.1.9.16.2.14', 'timeStampToken');
 
 // pkcs#12 safe bags
 _IN('1.2.840.113549.1.12.10.1.1', 'keyBag');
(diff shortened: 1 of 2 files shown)


Hatzl:
@davidlehn @dlongley is anything missing to merge this PR?

JulianAutogramm:
Would be really great to see this merged soon!

alexhaegler:
Great work! Would be awesome if this could be merged soon.

lib/oids.js
```diff
@@ -69,6 +69,7 @@ _IN('1.2.840.113549.1.9.14', 'extensionRequest');
 _IN('1.2.840.113549.1.9.20', 'friendlyName');
 _IN('1.2.840.113549.1.9.21', 'localKeyId');
 _IN('1.2.840.113549.1.9.22.1', 'x509Certificate');
+_IN('1.2.840.113549.1.9.16.2.14', 'timeStampToken');
```
davidlehn, line 72: This should be sorted up a few lines.  I assume this is useful for an unauthenticated attributes use case?
```suggestion
_IN('1.2.840.113549.1.9.16.2.14', 'timeStampToken');
_IN('1.2.840.113549.1.9.20', 'friendlyName');
_IN('1.2.840.113549.1.9.21', 'localKeyId');
_IN('1.2.840.113549.1.9.22.1', 'x509Certificate');
```

lib/pkcs7.js
```diff
@@ -245,6 +245,7 @@ p7.createSignedData = function() {
      *            (eg: forge.pki.oids.sha1).
      *          [authenticatedAttributes] an optional array of attributes
      *            to also sign along with the content.
+     *          [unauthenticatedAttributes] an optional array of unauthenticated attributes
```
davidlehn, line 248: Style fix.
```suggestion
     *          [unauthenticatedAttributes] an optional array of unauthenticated
     *            attributes.
```

lib/pkcs7.js
```diff
@@ -322,7 +323,7 @@ p7.createSignedData = function() {
         signatureAlgorithm: forge.pki.oids.rsaEncryption,
         signature: null,
         authenticatedAttributes: authenticatedAttributes,
-        unauthenticatedAttributes: []
+        unauthenticatedAttributes: signer.unauthenticatedAttributes || []
```
davidlehn, line 326: Does `unauthenticatedAttributes` need similar checks as above for `authenticatedAttributes`?

davidlehn (commented):
- Where is a good source to see how this would be used in practice?
- I really think this needs one or more tests to add some coverage to catch typos like that 'values' change.
- An short, or even partial, example in the docs would be good too.

dhensby:
not the OP but... unauthenticated attributes form part of the CMS standard (https://datatracker.ietf.org/doc/html/rfc5652#section-9.1)

In the unauthed attributes you can put data like what is defined here (https://datatracker.ietf.org/doc/html/rfc2985) and that's where the OIDs for "friendly name" etc have come from.

lib/oids.js
Strnadj: Fixed

lib/pkcs7.js
Strnadj: Fixed

lib/pkcs7.js
Strnadj: I am not sure, in `authenticatedAttributes` is at least one messageDigest attribute required... I assume the same condition won't be required for `unauthenticatedAttributes`, because messageDigest is already in `authenticatedAttributes`.

DoubleMalt:
Was this ever merged and implemented?

dhensby:
I don't believe so

PR #835, opened 2020-12-15
Added arch ppc64le to run on power systems

genisysram:
Hi
Here is my contribution to your code, its working good on powersystems.

Thanks for the code, its working good.
## What do these changes do?
Added Architecture "ppc64le"
## Are there changes in behavior for the user?
No

Added Architecure ppc64le to run on Powersystem

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+arch:
+    - amd64
+    - ppc64le
 language: node_js
 node_js:
   - "6"


davidlehn:
- I think all the code here is arch independent.  Is there a reason tests need to run on multiple archs?  What kind of issues were you trying to check for?  Endian issues?
- Are ppc64le test systems a limited resource that might cause test delays?
- Looks like it tried to build node 12 and timed out.
- Note we may also move over to GitHub actions at some point.  I'm unsure what kind of architecture support they have.

Issue #836, opened 2020-12-21
utf8string not being correctly parsed in forge.pki.certificateFromPem


brunovianarezende:
When I have an attribute whose type is utf8string, it isn't decoded as utf by forge.pki.certificateFromPem.  I've read the code and I see in asn1.js this comment:

/**
 * ASN.1 types. Not all types are supported by this implementation, only
 * those necessary to implement a simple PKI are implemented.
 */

so I wonder if it makes sense to create a patch to return decoded utf8 values.

If so, I tested locally and I just need to add one more conditional here: https://github.com/digitalbazaar/forge/blob/8753559e3881e4fe7db8bcc1d19003b2f5be7baa/lib/asn1.js#L617

PR #838, opened 2020-12-22
836: correctly decode utf8strings when loading a certificate from PEM

brunovianarezende:


brunovianarezende:
Closes #836

brunovianarezende:
I'm not so confident in this PR anymore. It seems like my test certificate authority was generating incorrect certificates when the subject had UTF-8 chars, when I fixed the test certificate authority, this patch caused utf-8 decode exceptions. However, I still need to understand why the test I created fail when the patch is not applied.

PR #855, opened 2021-03-14
Add OIDs for surname, title and givenName

renzenicolai:
The OIDs for surname, title and givenName were missing from the list in oids.js.

I need these OIDs to support certificates used by UZI pas cards. UZI pas is a means for Dutch healthcare professionals to identify themselves (see https://www.uziregister.nl/ and https://www.zorgcsp.nl/documents/RK1%20CPS%20UZI-register%20V10.2%20ENG.pdf). But these OIDs are very generic and probably usefull in many other cases.

This PR adds:
 - 2.5.4.4 surname
 - 2.5.4.12 title
 - 2.5.4.42 givenName

### DIFF
--- a/lib/oids.js
+++ b/lib/oids.js
@@ -104,6 +104,7 @@ _IN('2.16.840.1.101.3.4.1.42', 'aes256-CBC');
 
 // certificate issuer/subject OIDs
 _IN('2.5.4.3', 'commonName');
+_IN('2.5.4.4',  'surname');
 _IN('2.5.4.5', 'serialName');
 _IN('2.5.4.6', 'countryName');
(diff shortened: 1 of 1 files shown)


lib/oids.js
```diff
@@ -104,16 +104,19 @@ _IN('2.16.840.1.101.3.4.1.42', 'aes256-CBC');
 
 // certificate issuer/subject OIDs
 _IN('2.5.4.3', 'commonName');
+_IN('2.5.4.4',  'surname');
```
dhensby, line 107:
```suggestion
_IN('2.5.4.4', 'surname');
```

PR #860, opened 2021-03-25
Remove link to nodeguide.com

dhensby:
It appears nodeguide.com domain has been taken over by domain squatters and the style.html page appears to push a questionable PDF download.


davidlehn:
That's a shame.  Thanks for the patch.

PR #890, opened 2021-06-29
Rsapss csr

frk-pcn:
Pull request for RSA PSS padding to handle
- corrected calculation of salt length
- supported scheme in case of CSR signing

### DIFF
--- a/lib/x509.js
+++ b/lib/x509.js
@@ -683,7 +683,7 @@ var _readSignatureParameters = function(oid, obj, fillDefaults) {
   }
 
   if(capture.saltLength !== undefined) {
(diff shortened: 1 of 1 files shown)


davidlehn:
Do you have test cases for this change?
There are conflicts but it looks like it was rebased on main, not sure what's happening there.

frk-pcn:
I will see what's the best way to provide some tests to sign a CSR with RSA PSS padding. What do you think about changing tests/unit/csr.js to create a CSR in two flavors: with and without PSS padding?

frk-pcn:
I've enhanced the CSR test case. Unfortunately I'm not able to provide the change via github's pull request mechanism - currently no idea, why.

PR #896, opened 2021-07-28
fix compatibility in `usePureJavaScript` mode

boyxuper:
fix compatibility in `usePureJavaScript` mode, which globalScope can be undefined

### DIFF
--- a/lib/prng.js
+++ b/lib/prng.js
@@ -268,7 +268,7 @@ prng.create = function(plugin) {
     // use window.crypto.getRandomValues strong source of entropy if available
     var getRandomValues = null;
     var globalScope = forge.util.globalScope;
-    var _crypto = globalScope.crypto || globalScope.msCrypto;
+    var _crypto = globalScope !== undefined && (globalScope.crypto || globalScope.msCrypto);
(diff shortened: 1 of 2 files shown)


mattcollier:
Hello @boyxuper, thank you for the PR.  Can you determine how it is that `util.globalScope` is returning `undefined` in your use case?

https://github.com/digitalbazaar/forge/blob/master/lib/util.js#L118

boyxuper:
Yes, usePureJavaScript indicates a pure ES6 environment, which lack of `window` or `self` variables, and also not NodeJS runtime context.
It's exactly my production environment, it's `wechat miniapp` runtime context.

mattcollier:
Thanks for the additional detail.  It sounds like to me that `util.globalscope` is returning the value of `window` which is `undefined` in your case.

Based on this one line: https://github.com/digitalbazaar/forge/blob/master/lib/util.js#L123
```js
return typeof self === 'undefined' ? window : self;
```
I think adding the following conditional before L123 might work.
```js
// applicable for wechat miniapp runtime context
if(!(window || self)) {
  return {};
}
```

Which I believe would eliminate the need for the additional code changes you've made.

PR #900, opened 2021-08-20
🔒 change CSR examples to use 2048 bit keys

ctcpip:
2048 bits has been the required minimum since 2014


ctcpip:
tests passing btw

PR #911, opened 2021-09-10
Fix OID `serialName` to `serialNumber`.

davidlehn:
- OID 2.5.4.5 should be `serialNumber`.

@dlongley Was there a reason this id was using the wrong name?  Or was it just a typo?  What possible breakage could this cause?  Looking on the web, it seems the only places `serialName` comes up are places where people copied this code (oops!).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ Forge ChangeLog
 ### Added
 - OIDs for `surname`, `title`, and `givenName`.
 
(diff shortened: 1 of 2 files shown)


dlongley (approved):
Seems like a very long running typo to me. If we're worried about backwards compat, I think there's an mechanism for having aliases. Otherwise, I'm happy with the fix as is. Thanks!

PR #941, opened 2022-01-08
Add type to prime and prime.worker messages.

davidlehn:
- Add type to prime and prime.worker messages.
- Ignore unexpected messages.

I'm not sure how necessary this is, and it seems verbose.  But it will likely fix https://github.com/digitalbazaar/forge/pull/764 in a more robust way.  It could be optimized, but it's probably fine.


dlongley (approved):
Assuming we have test coverage for this and we're green, LGTM.

arnonrdp (approved):

PR #965, opened 2022-03-28
Allow DigestInfo.DigestAlgorith.parameters to be optional

dhensby:
Since v2.3.0 PKCS#1 signatures that don't include the `DigestAlgorithm.parameters` data do not parse successfully:

```
 errors: [
    '[DigestInfo.DigestAlgorithm] Tag class "0", type "16" expected value length "2", got "1"',
    '[DigestInfo] Tag class "0", type "16" expected value length "2", got "2"'
  ]
```

RFC3447 dictated that the `parameters` in the `DigestAlgorithm` were optional, however this was updated in RFC8017 to now need to be explicitly specified (even if `NULL`).

Prior to v2.3.0 of `forge` the PKCS#1 verify logic was permissive to the legacy format where the `parameters` were omitted (which was part of why I've been using it over Node's built in `verify`).

This change allows the parameters to, once again, be optional.

### DIFF
--- a/lib/rsa.js
+++ b/lib/rsa.js
@@ -286,6 +286,7 @@ var digestInfoValidator = {
       name: 'DigestInfo.DigestAlgorithm.parameters',
       tagClass: asn1.Class.UNIVERSAL,
       type: asn1.Type.NULL,
+      optional: true,
       constructed: false
     }]
   }, {


davidlehn:
Looks like based on the comments in the ASN.1 in both specs, NULL parameters are required for md2 and md5, but others are optional.  Fixed test and updated for md cases here: https://github.com/digitalbazaar/forge/pull/968.  Thanks!

dhensby:
Thanks for getting in this so promptly! 😍

PR #976, opened 2022-04-19
Test on Node.js 16.x and 18.x.

davidlehn:


davidlehn:
Yes, this is a lot of versions to test.  At some point ancient ones can be dropped.

PR #977, opened 2022-04-22
Support PEM BER data with unparsed trailing data.

davidlehn:
The hope is this less strict parsing doesn't reintroduce the issues fixed in 1.3.x.  I think this method of accepting unparsed trailing bytes for PEM data, still parsing by default as DER, and with the additional checks in 1.3.x, this should be ok.

### DIFF
--- a/lib/asn1.js
+++ b/lib/asn1.js
@@ -457,12 +457,53 @@ asn1.fromDer = function(bytes, options) {
   var byteCount = bytes.length();
   var value = _fromDer(bytes, bytes.length(), 0, options);
   if(options.parseAllBytes && bytes.length() !== 0) {
-    var error = new Error('Unparsed DER bytes remain after ASN.1 parsing.');
+    var error = new Error('Unparsed bytes remain after ASN.1 parsing.');
     error.byteCount = byteCount;
     error.remaining = bytes.length();
     throw error;
   }
   return value;
+};
+
+/**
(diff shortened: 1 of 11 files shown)


lib/asn1.js
```diff
@@ -457,14 +457,55 @@ asn1.fromDer = function(bytes, options) {
...
+ *            without schema support to understand the data context this can
+ *            erroneously decode values that happen to be valid ASN.1. This
+ *            flag will be deprecated or removed as soon as schema support is
+ *            available. (default: true)
+ *
+ * @throws Will throw an error for various malformed input conditions.
+ *
+ * @return the parsed asn1 object.
+ */
+asn1.fromPemBer = function(bytes, options) {
```
dlongley, line 496: This seems to break abstractions a little too much for my tastes. I understand the reasoning. I think that adding `asn1.fromBer` would be acceptable -- but we've always tried to avoid implementing an entire BER parser because it's usually not necessary, it's complicated, and a much bigger lift ... so that's not going to happen here.

Since this is PEM-specific, I would recommend that we add this utility only to `pem.js` instead. Perhaps renamed to `pem.toAsn1()` or `pem.berToAsn1()` or similar? Thoughts?

dlongley (changes requested):
I recommend we add this feature in a different spot -- otherwise, it looks ok.

Issue #952, opened 2022-02-17
Putty Private Key generated by SSH is not always valid.


yammelvin:
I wrote a test node js script to do the following:
1. generate a rsa key pair
2. using the generated rsa private key, generate the putty private key using forge
3. using the same generated rsa private key, generate the putty private key using the 'puttygen' command line.
4. compare the two putty private keys.

Out of every 100 times, about 7 to 10 times the ppk generated is not the same.
When it is not the same, the ppk by forge cannot be loaded or used by putty,
however the ppk by puttygen is ok.

Attached is the test script.

[testforgeputty.zip](https://github.com/digitalbazaar/forge/files/8091915/testforgeputty.zip)

mathieucaroff:
@yammelvin 

I've tried to run the test script, but I ran into issue with the puttygen parameters `-C` and `-O`

The documentation I've found for PuTTYgen does not mention these parameters either:

https://tartarus.org/~simon/putty-snapshots/htmldoc/Chapter8.html#puttygen-cli

> The options supported on the command line are:
> 
> -t keytype
>     Type of key to generate. You can select rsa, dsa, ecdsa, eddsa, ed25519, ed448, or rsa1. See [section 8.2.2](https://tartarus.org/~simon/putty-snapshots/htmldoc/Chapter8.html#puttygen-keytype). 
> -b bits
>     Size of the key to generate, in bits. See [section 8.2.3](https://tartarus.org/~simon/putty-snapshots/htmldoc/Chapter8.html#puttygen-strength). 
> --primes method
>     Method for generating prime numbers. You can select probable, proven, and proven-even. See [section 8.2.4](https://tartarus.org/~simon/putty-snapshots/htmldoc/Chapter8.html#puttygen-primes). 
> --strong-rsa
>     When generating an RSA key, make sure the prime factors of the key modulus are ‘strong primes’. See [section 8.2.4](https://tartarus.org/~simon/putty-snapshots/htmldoc/Chapter8.html#puttygen-primes). 
> --ppk-param key=value,...
>     Allows setting all the same details of the PPK save file format described in [section 8.2.13](https://tartarus.org/~simon/putty-snapshots/htmldoc/Chapter8.html#puttygen-save-params).

yammelvin:
@mathieucaroff

At the time of writing the test script.
I was using puttygen 0.76 which by default generates ppk version 2.
I just rerun the script using puttygen 0.78 which by default generates ppk version 3.
[...]

PR #987, opened 2022-06-15
bug fix privateKeyToPutty generation function

romeech91:
fix this problem: https://github.com/digitalbazaar/forge/issues/952


livestatio (approved):

yammelvin:
I ran the updated testscript with puttygen 0.78 against this fix.
All keys generated by puttygen and node-forge with this fix are identical.
See #952 for the updated testscript for puttygen 0.78

However, putty now uses ppk version 3 instead of ppk version 2, and node-forge is still generating ppk version 2.

PR #1025, opened 2023-03-28
:bug: implement ordered components of ASN.1 sets

### LINKED ISSUE #1013: ASN.1 `Set Of` invalid DER encoding for PKCS#7 authenticatedAttributes
According to the X.690 specification (ISO/IEC 8825-1:2021 (E)) the ASN.1 standard for `Set Of` components must be encoded in order when using DER:

>11 Restrictions on BER employed by both CER and DER
>...
>11.6 Set of components
>The encodings of the component values of a set-of value *shall appear in ascending order*, the encodings being compared as octet strings with the shorter components being padded at their trailing end with 0-octets. 
>NOTE – The padding octets are for comparison purposes only and do not appear in the encodings.

This has been determined to cause bouncy castle to reject the PKCS#7 signatures that this library produces.

I'll take a look into ordering the `Set Of` components, presumably this is best done in the `toDer` method on the asn1 package? Otherwise the sorting could be done when the ASN.1 is constructed.

dhensby:
Fixes #1013 

This implements sorting of ASN.1 types.

According to the X.690 specification (ISO/IEC 8825-1:2021 (E)) the ASN.1 standard for `Set Of` components must be encoded in order when using DER:

>11 Restrictions on BER employed by both CER and DER
>...
>11.6 Set of components
>The encodings of the component values of a set-of value *shall appear in ascending order*, the encodings being compared as octet strings with the shorter components being padded at their trailing end with 0-octets. 
>NOTE – The padding octets are for comparison purposes only and do not appear in the encodings.

This checks if the current ASN.1 object is a SET - if so it will order the encoded values. It also allows the `option` of `sort` to be set to force sorting, this is required when items need to be encoded as SET but have context specific tag class.

The ordering is done by comparing each byte in turn until we find either a difference or can determine they are the same.

Because of the change in how internals are now constructing the ASN.1 values, the PKCS#7 test output has had to change.

For completeness I have tested openssl is still able to validate the produced PKCS#7 signatures:

```
$ openssl smime -verify -attime 1332945777 -in ./signedDataWithAttrs1949GeneralizedTime-new.pem -signer certificate.pem -out signedtext.txt -inform PEM -CAfile certificate.pem
Verification successful
$ openssl smime -verify -attime 1332945777 -in ./signedDataWithAttrs1950UTCTime-new.pem -signer certificate.pem -out signedtext.txt -inform PEM -CAfile certificate.pem 
Verification successful
$ openssl smime -verify -attime 1332945777 -in ./signedDataWithAttrs2049UTCTime-new.pem -signer certificate.pem -out signedtext.txt -inform PEM -CAfile certificate.pem 
Verification successful
$openssl smime -verify -attime 1332945777 -in ./signedDataWithAttrs2050GeneralizedTime-new.pem -signer certificate.pem -out signedtext.txt -inform PEM -CAfile certificate.pem 
Verification successful
$ openssl smime -verify -attime 1332945777 -in ./detachedSignature-new.pem -signer certificate.pem -out signedtext.txt -inform PEM -CAfile certificate.pem -content content.txt
Verification successful
```

### DIFF
--- a/lib/asn1.js
+++ b/lib/asn1.js
@@ -188,6 +188,7 @@ asn1.Type = {
  * @param [options] the options to use:
  *          [bitStringContents] the plain BIT STRING content including padding
  *            byte.
+ *          [sort] sort the values
  *
  * @return the asn1 object.
  */
@@ -216,12 +217,17 @@ asn1.create = function(tagClass, type, constructed, value, options) {
     composed: constructed || forge.util.isArray(value),
     value: value
   };
-  if(options && 'bitStringContents' in options) {
-    // TODO: copy byte buffer if it's a buffer not a string
-    obj.bitStringContents = options.bitStringContents;
-    // TODO: add readonly flag to avoid this overhead
-    // save copy to detect changes
-    obj.original = asn1.copy(obj);
+  if(options) {
+    if('bitStringContents' in options) {
+      // TODO: copy byte buffer if it's a buffer not a string
+      obj.bitStringContents = options.bitStringContents;
+      // TODO: add readonly flag to avoid this overhead
+      // save copy to detect changes
+      obj.original = asn1.copy(obj);
+    }
+    if('sort' in options) {
+      obj.sort = options.sort || false;
+    }
   }
   return obj;
 };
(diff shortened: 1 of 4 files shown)


dhensby:
@davidlehn - sorry for another ping on this, but it would be really great to get this in

Issue #1028, opened 2023-04-06
 Cannot read property 'crypto' of undefined


xiuyu123123:
app.js错误:
 TypeError: Cannot read property 'crypto' of undefined
    at Object.t (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at e (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at Object.t (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at e (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at Object.t (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at e (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at Object.t (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at e (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at Object.t (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at e (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)(env: macOS,mp,1.06.2301171; lib: 2.30.3)

davidlehn:
More information is needed to help with this.

- What version of forge?
- What environment?
- What is the calling code?

There's some code that uses `globalScope.crypto`, which should work in modern runtimes.  Looks like there might be some older compatibility hacks from when `globalScope` was too new to use in `util.globalScope`.  I'm unsure what the state of that might be these days.  You might try code related to those spots to see what the issue is.

Issue #1027, opened 2023-04-06
Cannot read property 'crypto' of undefined


xiuyu123123:
app.js错误:
 TypeError: Cannot read property 'crypto' of undefined
    at Object.t (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at e (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at Object.t (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at e (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at Object.t (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at e (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at Object.t (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at e (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at Object.t (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)
    at e (vendor.js?t=wechat&s=1680776449375&v=64a94e62456c2f5e6e877651680b0c61:8952)(env: macOS,mp,1.06.2301171; lib: 2.30.3)

davidlehn:
Duplicate of https://github.com/digitalbazaar/forge/issues/1028

PR #1066, opened 2023-12-21
Add a fallback global scope for runtime likes wechat mini program

imyelo:
WeChat Mini Program is a popular application form in China, similar to Progressive Web Apps but with a specific runtime and system interface provided by the application platform (wechat indeeded). Its runtime is similar to a browser or React Native but lacks global objects like document, window, self. As a result, using node-forge in WeChat Mini Program would cause crashes due to the inability to find the global space, leading to issues like #1028 and #1027.

A few years ago, I made some minor efforts to support running such JavaScript modules in those environments, including [applying a runtime patch](https://github.com/tinajs/mina-webpack/blob/master/packages/mina-runtime-webpack-plugin/polyfill.js) for JS dependencies. However, this implementation heavily depends on webpack.

Especially with the popularity of various mini programs and cross-platform frameworks today, I believe that adding a fallback logic directly in node-forge can prevent crashes in such scenarios.

Hope this will be helpful for other mini program or hybrid developers as well. 😘

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -120,7 +120,11 @@ util.globalScope = (function() {
     return global;
   }
 
-  return typeof self === 'undefined' ? window : self;
+  return typeof self !== "undefined"
+    ? self
+    : typeof window !== "undefined"
+    ? window
+    : new Function("return this")();
 })();
 
 // define isArray


vricosti:
completely agree!!!
I have the same kind of issue.

PR #1082, opened 2024-05-22
fix: jsbn use util.isNodejs to determine node runtime

### LINKED ISSUE #1081: [jsbn] Node 21 introduced `navigator` global object which has changed jsbn behaviour
See https://github.com/digitalbazaar/forge/blob/v1.3.1/lib/jsbn.js#L120-L133

The JSBN library uses the global `navigator` object to make optimisation assumptions about the runtime. It is currently assumed that node has no `navigator` global object and, thus, will follow the `typeof navigator === 'undefined'` path. This is no longer the case in v21+ and so now a different path is taken.

For *now* this can be avoided by running node with the [`--no-experimental-global-navigator`](https://nodejs.org/docs/latest-v21.x/api/cli.html#--no-experimental-global-navigator) option, but presumably this will eventually become permanent and the optimisation logic will need to change.

dhensby:
fixes #1081 

A simple fix to avoid using the `navigator` global for determining if we are running node, instead use the forge `util.isNodejs` value.

TBH - I have no idea if the assumptions around performance in different browsers still holds since 2005 (almost 20 years ago) - but this at least keeps the node performance where it was.

### DIFF
--- a/lib/jsbn.js
+++ b/lib/jsbn.js
@@ -117,7 +117,7 @@ function am3(i,x,w,j,c,n) {
 }
 
 // node.js (no browser)
-if(typeof(navigator) === 'undefined')
+if(forge.util.isNodejs)
 {
    BigInteger.prototype.am = am3;
    dbits = 28;


davidlehn:
Did this come up because of an error or performance issue?  Or did you just happen to notice it?  I refreshed a PR to test up though 22 and tests pass without this patch.  Not sure if that path is tested though.
https://github.com/digitalbazaar/forge/pull/976

There is also `navigator` use in `util.estimateCores` and in `lib/random.js`.  Not sure if those paths would be hit in Node.js use.  A **lot** of code here needs refactoring to catch up with the last few decades of JS advancements.

As far as a minimal change that likely won't break anything, I think this patch will work.  Anyone else watching have thoughts?

dhensby:
This came up in a test we have in one of our projects for which compares the internals of a BigInteger to make sure our internal conversion of JWK to rsa keys performs as expected.

In Node 22 (and 21 from their changelog) there was a discrepancy because of this change.

I haven't done any testing of the actual performance implications, but it makes sense to continue to use 32 bit internals for node IMO.

I had found another slight improvement to JSBN off the back of this (when the number is clamped, it should truncate the `data` array, otherwise you can find times the array has vestigial elements at the end - this can easily be tested by providing a hex number with leading zeros), but I've left that out as it's a lot more "opinionated" change than just restoring the logic path for Node.

bolt-juri-gavshin:
Is there a chance for this PR to be merged (and released as 1.3.2) any time soon?

bolt-juri-gavshin:
@davidlehn, is there anything me (or community) can do to help release this?
It prevents migration to the latest NodeJS LTS (v22)

Neumann-Nils:
We would also like to see this merged, but seems like the package is inactive.
@davidlehn Are you looking for new maintainers for this repository?

Neumann-Nils:
@davidlehn I have seen that you have been more active on this repsoitory again, which is great 👍
Could you have a look here? We are currently need to fork this repository and apply the changes ourselves to make it work. We would love to see it being merged upstream.

PR #1085, opened 2024-06-05
Fix password mechanism

Sebastian-Dieguez:
Forge ChangeLog
===============

### Enhancement and interoperability
- **At [pkcs12.js] and [pbe.js].**
  - Fix in password mechanism.

### Enhancement and interoperability
- **At [pkcs12.js] and [pbe.js].**
  - Added (options.encryptCert) flag to allow encrypt [PKCS#7 ContentInfo], used in [PKCS#12 toPkcs12Asn1].
  - This ensures interoperability with [System.Security.Cryptography.X509Certificates.X509Certificate2(cert, password)].
  - Added suport for .NET interoperability for X509Certificate2(cert, password = "") 
	for do this set toPkcs12Asn1 with password = undefined.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,33 @@
 Forge ChangeLog
 ===============
 
+## 1.3.2 - 2024-06-04
+
+### Enhancement and interoperability
+- **At [pkcs12.js] and [pbe.js].**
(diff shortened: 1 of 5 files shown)


torstenwerner:
The change looks interesting. Is it planned to include this change in a released version any time soon?

PR #1106, opened 2025-01-03
Fix jsbn browser check

bolt-juri-gavshin:
fixes https://github.com/digitalbazaar/forge/issues/1081

Previous versions of NodeJS/Bun/Deno didn't have 'navigator' defined, while latest one do have it, which forces am2 algorithm with 26 dbits, which is awfully slow (~20x), at least in NodeJS v22 (compared to am3 with 28 dbits)

### DIFF
--- a/lib/jsbn.js
+++ b/lib/jsbn.js
@@ -116,8 +116,8 @@ function am3(i,x,w,j,c,n) {
   return c;
 }
 
-// node.js (no browser)
-if(typeof(navigator) === 'undefined')
+// node.js/bun/deno (no browser), latest versions have 'navigator' defined, but 'navigator.appName' is still undefined
+if(typeof(navigator) === 'undefined' || typeof(navigator.appName) === 'undefined')
 {
    BigInteger.prototype.am = am3;
    dbits = 28;


bolt-juri-gavshin:
@davidlehn this PR is an alternative to the [PR](https://github.com/digitalbazaar/forge/pull/1082) by @dhensby
Without the fix, upgrading from NodeJS v20 to v22 (LTS since end of 2024) requires nasty workarounds with global navigator (to continue performing).
Is it possible to get v1.3.2 with either of the fixes? Mine should cover other runtimes as well, but other one is also ok, as I care only about NodeJS :)

dhensby:
This feels like a bad idea, [appName](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/appName) is deprecated, so liable to be removed from browsers - meaning that it could cause unintended behaviour in browsers.

Also, we're still coupling with the `navigator` global and could still cause problems in the future if node/bun/deno/etc decide to populate this new property we now rely on to be `undefined`. The `utils.isNode` is specifically designed to determine if we are running in a node context, and thus makes more sense as a single-point-of-truth for this information and feature toggling.

bolt-juri-gavshin:
Thanks for the comment, I will close my PR then.
Is there a chance for the https://github.com/digitalbazaar/forge/pull/1082 to be merged/released any time soon?

Issue #552, opened 2017-12-18
AES-ECB decryption error - tmp.length is not a function


nathanschwarz:
Hi,
I'm trying to use the AES_ECB decryption mentionned in the doc.
the key is a string in a file and the encrypted data is also in an other file.
These were generated with openSsl.

forge keep throwing the same error which is 
`TypeError: tmp.length is not a function
    at forge.aes.Algorithm.initialize (path_to_app/node_modules/node-forge/lib/aes.js:203:19)
    at new forge.cipher.BlockCipher (path_to_app/node_modules/node-forge/lib/cipher.js:118:18)
    at Object.forge.cipher.createDecipher (path_to_app/node_modules/node-forge/lib/cipher.js:75:10)
at Object.<anonymous> (path_to_app/main.js:179:29)
`
and there's the code
`const forge = require('node-forge')
let aes_key = fs.readFileSync(AES_KEY_PATH, 'utf8')
let encrypted = fs.readFileSync('test/a.encrypted.jpg', 'utf8')
let decipher = forge.cipher.createDecipher('AES-ECB', aes_key)
decipher.mode.unpad = false;
decipher.start(); //line 179, where the error is thrown
decipher.update(forge.util.createBuffer(encrypted));
decipher.finish();
let decrypted = decipher.output.getBytes();
fs.writeFileSync('test/a.decrypted.jpg', decrypted)
`
Could you tell me what i'm doing wrong ? Thanks

ve3:
also happen with AES-CBC.

```
function decryptOpenssl(string, key) {
    string = window.atob(string);

    var md = forge.md.sha256.create();
    md.update(key);
    var hashKey = md.digest().toHex();
    var iv = hashKey.substr(0, 16);
    console.log(iv);

    // start error below.
    var decipher = forge.cipher.createDecipher('AES-CBC', key);
    decipher.start({'iv': iv});
    decipher.update(string);
    var result = decipher.finish();
    console.log(result);
    console.log(decipher.output.toHex());
}

decryptOpenssl('MDVlcDBvekhLYnlockJ4UTBkRm45YTlKQ2xNT1ZEL2NKbzV1czJCS0F3MD0=', 'sec_7Lk6$wMtzMbNoK%d&TgmCc5wh2oTq^');
```

should return `M66RA-GF0R2-A86WW-WF8Y3-PGW2W` but error `t.length`

davidlehn:
[...]

PR #1111, opened 2025-04-12
Fix Typo (tmp.length() -> tmp.length)

bigdale123:
So I think I encountered a typo in the aes.js lib file. On line 203, the length of `tmp` is found using `tmp.length()`. But,
    a) this results in a type error complaining that `tmp.length() is not a function` (See [Issue #552 ](https://github.com/digitalbazaar/forge/issues/552) 
    b) it is done the correct way (at least, as far as I can tell) not 10 lines earlier. I have provided an excerpt from `aes.js`:
```
183 |  if(typeof key === 'string' &&
184 |    (key.length === 16 || key.length === 24 || key.length === 32)) {
185 |    // convert key string into byte buffer
186 |    key = forge.util.createBuffer(key);
187 |  } else if(forge.util.isArray(key) &&
188 |    (key.length === 16 || key.length === 24 || key.length === 32)) {
189 |    // convert key integer array into byte buffer
190 |    tmp = key;
191 |    key = forge.util.createBuffer();
192 |    for(var i = 0; i < tmp.length; ++i) {
193 |      key.putByte(tmp[i]);
194 |    }
195 |  }
196 |
197 |  // convert key byte buffer into 32-bit integer array
198 |  if(!forge.util.isArray(key)) {
199 |    tmp = key;
200 |    key = [];
201 |
202 |    // key lengths of 16, 24, 32 bytes allowed
203 |    var len = tmp.length();
204 |    if(len === 16 || len === 24 || len === 32) {
205 |      len = len >>> 2;
206 |      for(var i = 0; i < len; ++i) {
207 |        key.push(tmp.getInt32());
208 |      }
209 |    }
210 |  }
```

So this pull request addresses that by chaning line 203 from `var len = tmp.length();` to `var len = tmp.length;`
Hope i'm not wrong about it, but it seems like a no brainer.

### DIFF
--- a/lib/aes.js
+++ b/lib/aes.js
@@ -200,7 +200,7 @@ forge.aes.Algorithm.prototype.initialize = function(options) {
     key = [];
 
     // key lengths of 16, 24, 32 bytes allowed
-    var len = tmp.length();
+    var len = tmp.length;
     if(len === 16 || len === 24 || len === 32) {
       len = len >>> 2;
       for(var i = 0; i < len; ++i) {


cyber-hari (approved):
## LGTM!

kendhia:
I came across this solution as I had same error but only once `node-forge` being used in a nextjs app is minified/bundled for production.
The minified code looked like:
```
                    c = [];
                    var e = (b = c).length();
                    if (16 === e || 24 === e || 32 === e) {
                        e >>>= 2;
                        for (var d = 0; d < e; ++d)
                            c.push(b.getInt32())
                    }
```

Which was throwing the error reported in this issue. But, the fix suggested causes another problem.
Since the minified code got `b=c`, b's length is always 0. So it will always fail!

kendhia:
This PR wasn't active for a while. So I created a new one https://github.com/digitalbazaar/forge/pull/1134

PR #1151, opened 2026-09-05
DigestInfo validation to prevent RSA-PKCS#1 v1.5 signature forgery

### LINKED ISSUE #1149: RSA PKCS#1 v1.5 signature forgery via garbage in DigestAlgorithm (incomplete CVE-2026-33894 fix)
reported via email on 4 June 2026 - no response.

The CVE-2026-33894 fix (af094e6) is incomplete. It only checks the top-level DigestInfo element count:

  lib/rsa.js:1174  if(!asn1.validate(obj, digestInfoValidator, capture, errors) || obj.value.length !== 2) { ... }

The nested DigestAlgorithm SEQUENCE has no element-count check, and asn1.validate ignores extra children. An attacker can stuff garbage bytes inside DigestAlgorithm; node-forge still accepts the DigestInfo and captures the attacker's digest. Because _parseAllDigestBytes defaults to true, trailing garbage is blocked, but garbage inside the structure is not.

I confirmed the parser gap on 1.4.0 (a DigestAlgorithm with [OID, NULL, garbage] validates, top-level length stays 2, attacker digest captured), and then forged a working signature end to end on an e=3 key with no private key:

  node-forge version: 1.4.0
  pub.verify(forged) => true        (message "forged-by-attacker-no-privkey")
  control (forged sig vs different message) => false

The forgery uses the standard low-exponent technique: the 32-byte digest is fixed at the bottom of the block, the ASN.1/padding framing at the top, and ~314 free "garbage" bytes inside DigestAlgorithm absorb the middle of S^3 so that S^3 is an exact cube matching a block node-forge accepts.

Suggested fix: require the DigestAlgorithm AlgorithmIdentifier SEQUENCE to contain exactly the OID plus optional NULL (add a nested value.length check), or make asn1.validate reject unconsumed children. Validating element counts at every nesting level (not just the outer SEQUENCE) closes the remaining slack.

Zelzazor:
Fixes #1149. 

Basically, validation of `DigestInfo` structure in `verify()` function for RSA PKCS#1 v1.5 in `rsa.js` was not validating any nested properties located between the required parts of said structure, allowing a forged signature to pass.

I've made an additional check inside the `if` of the `DigestInfo` structure validation, in order to check what is the length of the captured `DigestAlgorithm`, and compare it to it's only possible values.

A regression test was made inside `tests/unit/rsa.js` to detect the forged signature. Test was executed before making the change with an expected fail state. Then executed the full test suite: 816 passing, 0 failing.

Created a CHANGELOG entry documenting the fix and a CVE reference.

CVE: [CVE-2026-85393](https://www.cve.org/CVERecord?id=CVE-2026-85393)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,23 @@
 Forge ChangeLog
 ===============
 
+## 1.4.0 - 2026-xx-xx
+
+### Security
+- **HIGH**: Signature forgery in RSA-PKCS due to incomplete `DigestInfo` fix.
+  - The fix for CVE-2026-33894 in 1.4.0 only checked the element count of the
+    outer `DigestInfo` SEQUENCE. The nested `DigestAlgorithm` SEQUENCE was not
+    checked for extra, unconsumed elements, allowing the same class of low
+    public exponent (e.g. `e=3`) signature forgery via garbage bytes placed
+    inside that nested structure instead of the outer `DigestInfo` structure.
+  - Reported by geo-chen.
+  - CVE ID: [CVE-2026-85393](https://www.cve.org/CVERecord?id=CVE-2026-85393)
+
+### Fixed
+- [rsa] Fix RFC 8017 DigestInfo parsing to also require the nested
+  DigestAlgorithm sequence to contain only the OID and an optional NULL,
+  rejecting any extra, unconsumed elements. See #1149.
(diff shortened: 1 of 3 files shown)


Zelzazor:
Closed in favor of #1152

PR #1152, opened 2026-09-09
Fix RSA PKCS#1 v1.5 forgery via nested DigestAlgorithm garbage

### LINKED ISSUE #1149: RSA PKCS#1 v1.5 signature forgery via garbage in DigestAlgorithm (incomplete CVE-2026-33894 fix)

Krysthyan:
## Summary
- Fixes #1149 / [CVE-2026-85393](https://www.cve.org/CVERecord?id=CVE-2026-85393) (Snyk: SNYK-JS-NODEFORGE-19635204).
- The CVE-2026-33894 fix only enforced `obj.value.length === 2` on the outer `DigestInfo` SEQUENCE. `asn1.validate` still ignores extra children inside the nested `DigestAlgorithm` SEQUENCE, so attackers can stuff ~314 garbage bytes there and forge signatures for low public exponent keys (e.g. `e=3`).
- Require the nested `DigestAlgorithm` SEQUENCE to contain only the OID plus an optional NULL (same pattern as the outer length check).

## Test plan
- [x] Reproduction: forged EM with `DigestAlgorithm` children `[OID, NULL, garbage]` verified as `true` before the change; throws invalid DigestInfo after.
- [x] Added regression test `should check nested DigestAlgorithm element count` in `tests/unit/rsa.js`.
- [x] `npx mocha tests/unit/rsa.js` — 101 passing, 4 pending.
- [x] CHANGELOG Security + Fixed entries for 1.4.1.

Related: #1151 (same issue; this PR uses `obj.value[0].value.length` instead of adding `captureAsn1`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,23 @@
 Forge ChangeLog
 ===============
 
+## 1.4.1 - 2026-xx-xx
+
+### Security
+- **HIGH**: Signature forgery in RSA-PKCS due to incomplete `DigestInfo` fix.
+  - The fix for CVE-2026-33894 in 1.4.0 only checked the element count of the
+    outer `DigestInfo` SEQUENCE. The nested `DigestAlgorithm` SEQUENCE was not
+    checked for extra, unconsumed elements, allowing the same class of low
+    public exponent (e.g. `e=3`) signature forgery via garbage bytes placed
+    inside that nested structure instead of the outer `DigestInfo` structure.
+  - Reported by geo-chen.
+  - CVE ID: [CVE-2026-85393](https://www.cve.org/CVERecord?id=CVE-2026-85393)
+
+### Fixed
+- [rsa] Fix RFC 8017 DigestInfo parsing to also require the nested
+  DigestAlgorithm sequence to contain only the OID and an optional NULL,
+  rejecting any extra, unconsumed elements. See #1149.
+
 ## 1.4.0 - 2026-03-24
 
 ### Security
--- a/lib/rsa.js
+++ b/lib/rsa.js
@@ -1168,11 +1168,16 @@ pki.setRsaPublicKey = pki.rsa.setPublicKey = function(n, e) {
             parseAllBytes: options._parseAllDigestBytes
           });
 
-          // validate DigestInfo structure and element count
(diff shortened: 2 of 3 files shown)


Zelzazor:
Hi @Krysthyan 

I can get behind this solution instead of the one I've made, since it's simpler and does not introduce the potential overhead of adding captureAsn1 to check the same thing.

Have you run the rest of the suites just in case they also pass?

@davidlehn can you check this? The CVE is blocking several pipelines as much as I can see in the mentions of #1149

I will close my PR in favor of this one.

gwenael-louandre:
Would it be possible to review this PR? We have a vulnerability on our application due to it.
image

